PR #13, opened 2024-09-12
Feature/Add method to validate environment in a checkpoint

HCookie:
Add `validate_environment` to Checkpoint 
- Either warn the user or raise exception


HCookie:
Strange issue I need to look into tomorrow.
The `gather_provenance_info` seems to only detect imported modules, so when run on a checkpoint it complains about missing `anemoi-models`, but then when `anemoi-models` is successfully imported it is captured.

b8raoult (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Keep it human-readable, your future self will thank you!
 ### Added
 - ci: changelog release updater
 - earthkit-data replaces climetlab
+- `validate_environment` on Checkpoint
```
mchantry, line 16: Please can you link to the PR at the end of the link ie #13 thanks

src/anemoi/inference/checkpoint/__init__.py
```diff
@@ -63,3 +65,98 @@ def operational_config(self):
...
+
+        for module in train_environment["module_versions"].keys():
+            if not all_packages and "anemoi" not in module:
+                continue
+
+            if module not in inference_environment["module_versions"]:
+                invalid_messages["missing"].append(f"Missing module in inference environment: {module}")
+            elif train_environment["module_versions"][module] != inference_environment["module_versions"][module]:
+                invalid_messages["mismatch"].append(
+                    f"Version mismatch for module {module}: {train_environment['module_versions'][module]} != {inference_environment['module_versions'][module]}"
```
gmertes, line 119: Can you update the log message to make it clear which one is the training version, and which one the inference env version. Same for the sha1 log

src/anemoi/inference/checkpoint/__init__.py
```diff
@@ -63,3 +65,98 @@ def operational_config(self):
...
+                f"Python version mismatch: {train_environment['python']} != {inference_environment['python']}"
+            )
+
+        for module in train_environment["module_versions"].keys():
+            if not all_packages and "anemoi" not in module:
+                continue
+
+            if module not in inference_environment["module_versions"]:
+                invalid_messages["missing"].append(f"Missing module in inference environment: {module}")
+            elif train_environment["module_versions"][module] != inference_environment["module_versions"][module]:
```
gmertes, line 117: For most modules it will be fine if version inference > training. So I would split this case up with different log messages so it's immediately clear if it's a problem or not:
- inference > training : a gentle reminder, but probably not a problem
- inference < training : danger, this is a problem

gmertes (changes requested):
Nice!

One other thing to consider: there will be modules that are present during training, but are not required during inference (dependencies specified in anemoi-training that are not in anemoi-inference)

For example:
anemoi-training itself, anemoi.datasets, hydra, hydra_plugins, pytorch-lightning

For these obvious ones, should we maybe hard-code some exceptions? Because it's not useful for the validation to tell me anemoi-training is missing in an inference environment.

src/anemoi/inference/checkpoint/__init__.py
```diff
@@ -63,3 +65,98 @@ def operational_config(self):
 
         LOG.warning("No operational configuration found. Using default configuration.")
         return {}
+
+    def validate_environment(
+        self,
+        all_packages: bool = False,
+        on_difference: Literal["warn", "error"] = "warn",
+    ) -> int:
```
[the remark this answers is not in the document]
HCookie: Changed in next commit

src/anemoi/inference/checkpoint/__init__.py
HCookie: Changed in next commit

src/anemoi/inference/checkpoint/__init__.py
```diff
@@ -63,3 +67,125 @@ def operational_config(self):
...
+
+        if train_environment["python"] != inference_environment["python"]:
+            invalid_messages["python"].append(
+                f"Python version mismatch: {train_environment['python']} != {inference_environment['python']}"
+            )
+
+        for module in train_environment["module_versions"].keys():
+            if not all_packages and "anemoi" not in module:
+                continue
+            elif module.split(".")[0] in exempt_packages:
```
gmertes, line 133: The exemption doesn't work on the anemois, you need another case for namespace packages. 

In `EXEMPT_PACKAGES` you have the anemois with a dash instead of a dot.

src/anemoi/inference/checkpoint/__init__.py
```diff
@@ -10,11 +10,15 @@
...
 from functools import cached_property
+from typing import Literal
 
 from anemoi.utils.checkpoints import has_metadata
 from anemoi.utils.checkpoints import load_metadata
+from anemoi.utils.provenance import gather_provenance_info
 
 from .metadata import Metadata
+from .package_exemptions import EXEMPT_PACKAGES
+from .utils import Version
```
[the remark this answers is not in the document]
HCookie: Addressed in upcoming commit

src/anemoi/inference/checkpoint/__init__.py
HCookie: Addressed in upcoming commit

gmertes (approved):

PR #15, opened 2024-09-19
Feature: Add Condition

### LINKED ISSUE #16: Add Data Management Class
Add a class to store data

Should be instantiable from xarray, earthkit, numpy

Runners should then use this to run data autorecurrently

HCookie:
- Stores data as a dictionary
- Can be made from xarray, earthkit, numpy, or dict
- Can be exported to array

Closes #16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Keep it human-readable, your future self will thank you!
 - ci: changelog release updater
 - earthkit-data replaces climetlab
 - ci-hpc-config
+- Add Condition to store data
 
 ### Changed
 - ci: updated workflows on PR and releases to use reusable actions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Keep it human-readable, your future self will thank you!
 - ci: changelog release updater
 - earthkit-data replaces climetlab
 - ci-hpc-config
-- Add Condition to store data
+- Add Condition to store data [#15](https://github.com/ecmwf/anemoi-inference/pull/15)
 
 ### Changed
 - ci: updated workflows on PR and releases to use reusable actions
--- a/src/anemoi/inference/condition.py
+++ b/src/anemoi/inference/condition.py
@@ -8,10 +8,10 @@ from typing import Callable
 from typing import TypeVar
 
 import numpy as np
-import xarray as xr
 
 if TYPE_CHECKING:
     import earthkit.data as ekd
+    import xarray as xr
 
 Array = TypeVar("Array")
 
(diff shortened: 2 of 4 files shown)


b8raoult (approved):

src/anemoi/inference/condition.py
```diff
@@ -0,0 +1,285 @@
...
+        data = data.transpose(variable_dim, *dims)
+
+        # Get all variables and their values in the data
+        variable_dict = {}
+        for var in data.coords[variable_dim].values:
+            variable_dict[var] = data.sel({variable_dim: var}).values
+        return Condition(variable_dict, private_info=private_info)
+
+    @classmethod
+    def from_earthkit(self, fieldlist: "ekd.FieldList", private_info: Any = None, **kwargs) -> Self:
```
gmertes, line 208: I like `-> Self` but it is a python >= 3.11 feature, and our minimum supported version is 3.9

So instead, please do `from __future__ import annotations` and return the class name explicitly.

tests/test_condition.py
```diff
@@ -0,0 +1,39 @@
...
+from hypothesis import assume
+from hypothesis import given
+from hypothesis import strategies as st
+from hypothesis.extra import numpy as npst
+
+from anemoi.inference.condition import Condition
+
+
+@given(shape=npst.array_shapes(min_dims=2), data=st.data())
+def test_code(shape, data):
```
gmertes, line 11: Can you rename this to `test_from_numpy` or something more explicit

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Keep it human-readable, your future self will thank you!
 - ci: changelog release updater
 - earthkit-data replaces climetlab
 - ci-hpc-config
+- Add Condition to store data
```
gmertes, line 17: link to PR

src/anemoi/inference/condition.py
```diff
@@ -0,0 +1,285 @@
+import itertools
+import re
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Callable
+from typing import Self
+from typing import TypeVar
+
+import numpy as np
+import xarray as xr
```
gmertes, line 10: The xarray import is not lightweight (on Atos it takes about 30s), and since we don't use it by default, should we put the import where it is used (so in from_xarray) ?

gmertes (changes requested):
This is great work! Just some minor comments.

src/anemoi/inference/condition.py
b8raoult: We should also not force people to install xarray to use `anemoi-inference` if they never need it. So, yes, we import in the method itself.

src/anemoi/inference/condition.py
HCookie: Apologies, addressed in upcoming commit

src/anemoi/inference/condition.py
HCookie: Sure, I wasn't clear on thoughts around delayed imports as some linters really don't like it.
Fixed in upcoming commit

gmertes (approved):

PR #22, opened 2024-09-27
Adjustments required as part of ECMWF ML PP WP1 Task 1.3

MeraX:
These changes are required in order to prepare anemoi for ICON data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@ Keep it human-readable, your future self will thank you!
 - Add support for unstructured grids
 
 ### Changed
-- use `PS` if `lsm` is not available as prognostic template (#23)
 
 - Add cos_solar_zenith_angle to list of known forcings
 - Add missing classes in checkpoint handling
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Keep it human-readable, your future self will thank you!
 ### Added
 
 ### Changed
+- use `PS` if `lsm` is not available as prognostic template
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Keep it human-readable, your future self will thank you!
 - Add support for unstructured grids
 
 ### Changed
+- Use ICON grid in inference plugin runner
 
 - Add cos_solar_zenith_angle to list of known forcings
 - Add missing classes in checkpoint handling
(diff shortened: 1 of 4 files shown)


src/anemoi/inference/runner.py
```diff
@@ -312,7 +313,10 @@ def get_most_recent_datetime(input_fields):
 
         most_recent_datetime = get_most_recent_datetime(input_fields)
         reference_fields = [f for f in input_fields if f.datetime()["valid_time"] == most_recent_datetime]
-        prognostic_template = reference_fields[self.checkpoint.variable_to_index["lsm"]]
+        if "lsm" in self.checkpoint.variable_to_index:
+            prognostic_template = reference_fields[self.checkpoint.variable_to_index["lsm"]]
+        else:
+            prognostic_template = reference_fields[self.checkpoint.variable_to_index["PS"]]
```
mchantry, line 319: I think it would be better to have a list of possible variables that can be used as a template.
The list could be defined as input variable to run, with default list of possibilities, then this could be further generalised in the future.
`prognostic_template_variable_list=["lsm","PS","sp"]`

src/anemoi/inference/checkpoint/metadata/__init__.py
```diff
@@ -115,6 +115,22 @@ def hour_steps(self):
 
         return int(n[:-1]) * {"h": 1, "d": 24}[n[-1]]
 
+    @cached_property
+    def max_refinement_level_c(self):
```
mchantry, line 119: Where is this used? Will this come in another PR?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Keep it human-readable, your future self will thank you!
 ### Added
```
mchantry, line 13: `max_refinement_level_c` does this need to be reference here?

HCookie (changes requested):
Please undo the changes to the copyright statements,
A general solution to contributions will be propsed and included soon

src/anemoi/inference/runner.py
MeraX: #29 made this change obsolete.

src/anemoi/inference/checkpoint/metadata/__init__.py
MeraX: I added the code that uses this method

MeraX:
I made a new push with the setup that we are currently using to do inference with our unstructured ICON grid. At this stage the PR is rather a preview than an actual PR.

PR #23, opened 2024-09-27
Enable inference when no constant forcings are used

MeraX:
Inference failed when no constant forcings were used in the training.

The error message related to the fact, that an empty list in `computed_forcing_mask` can't be used as an index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Keep it human-readable, your future self will thank you!
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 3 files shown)


HCookie (changes requested):
Please undo the changes to the copyright statements,
A general solution to contributions will be propsed and included soon

HCookie:
> Thanks for this note. Do you already know when and how this will happen?
It is currently being discussed by the ATS. I cannot give a timeframe at this moment

PR #39, opened 2024-10-30
fix dataset path

dietervdb-meteo:
The calling of the dataset path was broken, since .zarr was stripped. Furthermore it is more convenient to keep the full dataset path as set in the checkpoint as the default. In the fix here the path as read from the checkpoint is directly fed to anemoi-datasets `open_dataset`.

### DIFF
--- a/src/anemoi/inference/inputs/dataset.py
+++ b/src/anemoi/inference/inputs/dataset.py
@@ -9,7 +9,6 @@
 
 
 import logging
-import os
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


dietervdb-meteo:
Stripping of .zarr was actually implemented for those users that provide ~/.config/anemoi/settings.toml. This fix now restores support for that option but also provides support for loading the dataset from the complete path in case of a configuration without ~/.config/anemoi/settings.toml.

Issue #34, opened 2024-10-24
LAM rollout


gmertes:
When doing inference with LAM, the model will output an interior prediction + the boundary conditions. 

During rollout, when we feed the output back to the input, we need a way to replace the boundary conditions with new initial condition at the next timestep (e.g.: ERA5). 

We could treat this boundary layer as a forcing: input fields that get updated at every tilmestep from disk. Using the cutout as a mask.

dietervdb-meteo:
The branch feature/refactor currently has an implementation of this, as a subset of forcings referred to as 'boundary forcing'. It is currently automatically turned on/off if `output_mask: cutout` is persent or not in the model configuration as saved in the checkpoint. For the moment the implementation only supports inference from the same [cutout] dataset as used during training.

PR #42, opened 2024-11-04
add boundary forcings for lam

dietervdb-meteo:
This is a first attempt to address issue  LAM rollout #34 . 

It appears to be technically working [when running from a dataset]. Needs to be checked if indeed everything is done correctly. Please do not merge as is.


b8raoult (approved):

dietervdb-meteo:
Did some detailed testing of the boundary forcing as implemented by this PR. It seems to work.

PR #55, opened 2024-11-26
add model parallel for inference

cathalobrien:
Lets you run inference over multiple GPUs. 

All credit goes to @mishooax . This is his implementation, I just added it to anemoi inference

I compared output for n320 1024c running on 1 GPU vs 4 GPUs and it seems to match.

With this I was able to run 9km inference over 4 nodes with 4 40GB a100s per node. 

I would like feedback about how the input tensor is read and how the output tensor is written. Currently all ranks read the input and only rank 0 writes output. Also, at the moment when you run there is lots of duplicated logging 

Unfortunately you have to use slurm to launch an inference job on multiple GPUs (as opposed to anemoi training which supports launching interactive jobs with multiple gpus like `anemoi-training train hardware.num_gpus_per_node=<num_gpus>`). I tried launching with 'torchrun' but it didnt work. Happy to look into this more though

If you are running over multiple nodes you need to add these lines to your slurm batch script
```bash
MASTER_ADDR=$(scontrol show hostname $SLURM_NODELIST | head -n 1)
export MASTER_ADDR=$(nslookup $MASTER_ADDR | grep -oP '(?<=Address: ).*')
```
If you're running over a single node, `localhost` is used as the address.

### DIFF
--- a/src/anemoi/inference/runner.py
+++ b/src/anemoi/inference/runner.py
@@ -14,7 +14,11 @@ from functools import cached_property
 
 import numpy as np
 import torch
+import torch.distributed as dist
 from anemoi.utils.timer import Timer
+
+
+import os
 
 from .checkpoint import Checkpoint
 
@@ -375,6 +379,28 @@ class Runner:
                         model_index,
                     )
                     raise ValueError(f"Field '{name}' has NaNs and is not marked as imputable")
+                
+                
+        global_rank = int(os.environ["SLURM_PROCID"])  # Get rank of the current process, equivalent to dist.get_rank()
+        LOGGER.info("Global rank: %d", global_rank)
+        world_size = int(os.environ["SLURM_NTASKS"])  # Total number of processes
+        LOGGER.info("World size: %d", world_size)
+        dist.init_process_group(
+              backend="nccl",
+              init_method=f'tcp://{os.environ["MASTER_ADDR"]}:{os.environ["MASTER_PORT"]}',
+              timeout=datetime.timedelta(minutes=3),
+              world_size=world_size,
+              rank=global_rank,
+            )
+        
+        local_rank = int(os.environ.get("SLURM_LOCALID", 0))
+        LOGGER.info("Global rank %d has also local rank: %d ...", global_rank, local_rank)
+        
+        # Ensure each process only uses one GPU
+        torch.cuda.set_device(local_rank)
+        
+        model_comm_group_ranks = np.arange(world_size, dtype=int)
+        model_comm_group  = torch.distributed.new_group(model_comm_group_ranks)
 
         with Timer(f"Loading {self.checkpoint}"):
             try:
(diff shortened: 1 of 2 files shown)


mishooax:
thanks @cathalobrien - this looks good, just a minor comment from my side.
sadly i don't have the time to test this - if someone else wants to, please go ahead.

src/anemoi/inference/runner.py
```diff
@@ -376,6 +379,32 @@ def run(
...
+        world_size = int(os.getenv("SLURM_NTASKS", 1))  # Total number of processes
+        local_rank = int(os.getenv("SLURM_LOCALID", 0))  # Get GPU num of current process
+        if global_rank == 0:
+            LOGGER.info("World size: %d", world_size)
+        addr = os.getenv("MASTER_ADDR", "localhost")  # localhost should be sufficient to run on a single node
+        port = os.getenv("MASTER_PORT", 10000 + random.randint(0, 10000))  # random port between 10,000 and 20,000
+        dist.init_process_group(
+            backend="nccl",
+            init_method=f"tcp://{addr}:{port}",
+            timeout=datetime.timedelta(minutes=1),
```
mishooax, line 394: we should probably set a longer timeout, O(5-10 mins) ?

src/anemoi/inference/runner.py
```diff
@@ -443,55 +474,57 @@ def get_most_recent_datetime(input_fields):
 
             # Predict next state of atmosphere
             with torch.autocast(device_type=device, dtype=autocast):
-                y_pred = model.predict_step(input_tensor_torch)
+                # y_pred = model.predict_step(input_tensor_torch, model_comm_group)
+                y_pred = model.forward(input_tensor_torch.unsqueeze(2), model_comm_group)
```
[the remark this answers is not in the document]
cathalobrien: It expects an ensemble dimension I believe?

src/anemoi/inference/runner.py
cathalobrien: yeah seems sensible

src/anemoi/inference/runner.py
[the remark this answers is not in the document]
mishooax: ah, yes, somehow i missed that you are now calling forward instead of predict_step.
is predict_step now obsolete? (if so, should it be removed?)

src/anemoi/inference/runner.py
```diff
@@ -376,6 +379,32 @@ def run(
...
 
+        global_rank = int(
+            os.getenv("SLURM_PROCID", 0)
+        )  # Get rank of the current process, equivalent to dist.get_rank()
+        world_size = int(os.getenv("SLURM_NTASKS", 1))  # Total number of processes
+        local_rank = int(os.getenv("SLURM_LOCALID", 0))  # Get GPU num of current process
+        if global_rank == 0:
+            LOGGER.info("World size: %d", world_size)
+        addr = os.getenv("MASTER_ADDR", "localhost")  # localhost should be sufficient to run on a single node
+        port = os.getenv("MASTER_PORT", 10000 + random.randint(0, 10000))  # random port between 10,000 and 20,000
```
japols, line 390: Does the fallback to `10000 + random.randint(0, 10000)` work without consistent seeding across ranks? Maybe we should use `SLURM_JOBID` instead?

b8raoult:
I'll close this one because it is now out of date.

PR #58, opened 2024-11-26
Move validate_environment to metadata

HCookie:


src/anemoi/inference/metadata.py
```diff
@@ -457,6 +476,145 @@ def _print(title, provenance):
         LOG.warning("If you are running from a git repository, the versions reported above may not be accurate.")
         LOG.warning("The versions are only updated after a `pip install -e .`")
 
+    def validate_environment(
```
theissenhelen, line 479: I know this a simple refactor and just moving code around. But this is a really long method covering more than 100 lines of code. It contains elements that could be factored out into their own methods to improve readability.

theissenhelen (changes requested):

src/anemoi/inference/metadata.py
HCookie: Addressed in [97df5fe](https://github.com/ecmwf/anemoi-inference/pull/58/commits/97df5fef29d9bab4d99993cf39685a50404646e6)

b8raoult (approved):

b8raoult (approved):

Issue #48, opened 2024-11-18
Improve raw output.


dietervdb-meteo:
### Is your feature request related to a problem? Please describe.

The branch feature/refactor supports various types of output. One of them is 'raw' output, which simply saves the states in a format essentially identical to how they are used internally, i.e. a set of labelled numpy arrays.

1. Currently the naming of the output files is hard coded and supports only hourly frequency. This should be updated to support at least minute level frequency. Furthermore a more uniform naming among different output files should be the aim. For example the naming of the raw output files and the plot outputs should converge.

2. (Part of) the input state is also saved in the output. For raw output the first state currently as only one datetime stamp but contains the numpy arrays of both $$t_0$$ and $$t_{-1}$$. Either only $$t_0$$ should be saved as output or both could be part of the output but then they should be saved as separate files with different datetime stamps.

 

### Organisation

RMI

PR #62, opened 2024-11-28
Improved output of initial state

### LINKED ISSUE #61: Add initial state to netcdf output.

Currently writing the initial state to netcdf output is not supported.

_No response_

_No response_

### Additional context

_No response_

### Organisation

RMI

dietervdb-meteo:
This PR closes #61 and #48. 
A new function `reduce()` is added to the `Output` class, which projects a state to the last step in the multi-step dimension.

This function is then used to define `write_initial_state()` as a combination of `reduce()` and `write_state()` for the netcdf, raw and plot output.

The naming of the raw output is brought in line with the naming of the plot output.

The default `write_initial_state` config setting is now `true`.

----
📚 Documentation preview 📚: https://anemoi-inference--62.org.readthedocs.build/en/62/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,9 @@ Keep it human-readable, your future self will thank you!
 - Add CONTRIBUTORS.md file (#36)
 
 ### Changed
+- Change `write_initial_state` default value to `true`
 - Raw output of initial state contains only values at initial time
-- Changed default naming of raw output
+- Change default naming of raw output
 - Add cos_solar_zenith_angle to list of known forcings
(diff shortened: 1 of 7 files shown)


dietervdb-meteo:
Questions:
1. Currently default config is not to write the intial state to output. Shouldn't it be better if the default is to write the initial state?
2. Plot output does the projection to last step of in the multi-step dimension explicitly (and for all states, not just the initial state). This could now be done using `reduce()` and possibly applied only in `write_initial_state`. Should we change this?

b8raoult (dismissed):

PR #73, opened 2024-12-05
Provide options to patch the checkpoint metadata when running a model

b8raoult:


src/anemoi/inference/commands/metadata.py
```diff
@@ -91,22 +100,31 @@ def add_arguments(self, command_parser):
             default=os.environ.get("EDITOR", "vi"),
         )
 
+        command_parser.add_argument(
+            "--pager",
+            help="Editor to use for the ``--view`` option. Default to ``$PAGER`` if defined, else ``less``.",
+            default=os.environ.get("PAGER", "less"),
```
HCookie, line 106: I presume PAGER is the standard env var for editors?

HCookie (dismissed):

src/anemoi/inference/commands/metadata.py
b8raoult: Yes

PR #74, opened 2024-12-05
Output archive requests json with no indent by default

gmertes:
Ouput the archive json with no indent by default, so it is only on 1 line. This makes it easier to handle the file afterwards, like concat into a jsonl file.


mchantry (approved):
LGTM nice work.

PR #88, opened 2024-12-18
feat(retrieve): Use input grib as target

gmertes:
If the config contains an input grib, the retrieve command will use that are the target in the generated mars request. 
Also make the default output stdout.

config.yaml:
```yaml
input:
  grib: input.grib
checkpoint: model.ckpt
```

```
anemoi-inference retrieve config.yaml --date 20240101T00:00:00
```
will have target=input.grib

Target can still be overwritten by passing `--extra target=anotherinput.grib`


b8raoult (commented):
What is the purpose here?

PR #100, opened 2025-01-08
ci: turn off hpc workflow

theissenhelen:
The HPC runners do not come with GPUs and are therefore redundant for now.


gmertes (approved):
LGTM

We can consider moving the HPC runners to a weekly task or on-demand.

PR #104, opened 2025-01-13
Turn on python 3.9 QA and fix typing

gmertes:
- Turn on python 3.9 in the QA version matrix
- `type | None` is unsupported  in python 3.9, replace it with `Optional[type]` or `Union[type, type, None]`

Other instances of `|` do work thanks to `__future__.annotations` , so I left those alone.


HCookie:
Given we use the `import __future__.annotations` in places, why was the decision made to change the `|` erroring types and not just add the import?

HCookie (approved):

PR #106, opened 2025-01-14
Instantiate runner from config and add CRPS runner

gmertes:
To support running the CRPS model, which has an extra argument `fcstep` in the `predict_step` fn. This argument is only picked up by the new `CrpsRunner`, and ignored by default.

Any other future extra arguments can be passed in this way, and only used by a specific runner.

### DIFF
--- a/src/anemoi/inference/commands/run.py
+++ b/src/anemoi/inference/commands/run.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 import logging
 
 from ..config import load_config
-from ..runners.default import DefaultRunner
+from ..runners import create_runner
 from . import Command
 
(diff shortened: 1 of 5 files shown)


b8raoult (approved):

HCookie:
Is adding a runner for each different model like this sustainable? Should we add a plugin approach to allow for external extension?

gmertes:
> Is adding a runner for each different model like this sustainable? Should we add a plugin approach to allow for external extension?

Indeed, but I don't plan on having a new runner for every model, only if the interface changes like it did in this case.

Either we do a plugin approach, or a more generic way of passing options to models.

PR #108, opened 2025-01-16
feat: Added the option to run inference in parallel

cathalobrien:
This PR allows you to run inference across multiple GPUs and nodes. 

Parallel inference relies on [PR77](https://github.com/ecmwf/anemoi-core/pull/77) to anemoi-models. Running sequentially with an older version of models will still work. Trying to run in parallel with an older version will prompt you to upgrade your anemoi models. 
When a newer version of models is released with PR77 included, It might be worthwhile bumping the minimum version required by anemoi-inference.

Currently Slurm is required to launch the parallel processes, and some Slurm env vars are read to set up the networking. Below is an example Slurm batch script 
```bash
#!/bin/bash
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=4
#SBATCH --gpus-per-node=4
#SBATCH --cpus-per-task=8
#SBATCH --time=0:05:00
#SBATCH --output=outputs/paralell_inf.%j.out

source env.sh
srun anemoi-inference run inf.yaml
```

One QOL of life feature would be make only process 0 log. Currently you get a lot of spam when running in parallel because each process logs. Any ideas how to do this would nicely would be great
```
2025-01-16 08:22:53 INFO Accumulating fields ['cp', 'tp']
2025-01-16 08:22:53 INFO Accumulating fields ['cp', 'tp']
2025-01-16 08:22:53 INFO Accumulating fields ['cp', 'tp']
2025-01-16 08:22:53 INFO Accumulating fields ['cp', 'tp']
...
2025-01-16 08:23:26 INFO World size: 4
```

----
📚 Documentation preview 📚: https://anemoi-inference--108.org.readthedocs.build/en/108/


src/anemoi/inference/parallel.py
```diff
@@ -0,0 +1,79 @@
+import datetime
```
HCookie, line 1: Licence statement needed here

src/anemoi/inference/runner.py
```diff
@@ -239,25 +241,38 @@ def model(self):
             return model
 
     def forecast(self, lead_time, input_tensor_numpy, input_state):
+
+        # determine processes rank for parallel inference and assign a device
+        global_rank, local_rank, world_size = get_parallel_info()
+        self.device = f"{self.device}:{local_rank}"
+        torch.cuda.set_device(local_rank)
```
HCookie, line 248: We should be mindful of locking to cuda, but that is likely out of scope for this PR

src/anemoi/inference/parallel.py
```diff
@@ -0,0 +1,79 @@
...
+        model_comm_group = None
+
+    return model_comm_group
+
+
+def get_parallel_info():
+    """Reads Slurm env vars, if they exist, to determine if inference is running in parallel"""
+    local_rank = int(os.environ.get("SLURM_LOCALID", 0))  # Rank within a node, between 0 and num_gpus
+    global_rank = int(os.environ.get("SLURM_PROCID", 0))  # Rank within all nodes
+    world_size = int(os.environ.get("SLURM_NTASKS", 1))  # Total number of processes
```
HCookie, line 77: I'm not sure I like the only automatic approach, a user should have some control to turn off parallelisation even if in a slurm job.

src/anemoi/inference/parallel.py
```diff
@@ -0,0 +1,79 @@
...
+
+def init_parallel(global_rank, world_size):
+    """Creates a model communication group to be used for parallel inference"""
+
+    if world_size > 1:
+
+        master_addr, master_port = init_network()
+        dist.init_process_group(
+            backend="nccl",
+            init_method=f"tcp://{master_addr}:{master_port}",
```
HCookie, line 59: Are other communication protocols supported?

src/anemoi/inference/parallel.py
```diff
@@ -0,0 +1,79 @@
...
+
+import numpy as np
+import torch
+import torch.distributed as dist
+
+LOG = logging.getLogger(__name__)
+
+
+def init_network():
+    """Reads Slurm environment to set master address and port for parallel communication"""
```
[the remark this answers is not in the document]
tmi: ftr, it's not a hypothetical question, I do exactly that in Cascade :) And we have two layers of the problem:
- how to allow Amenoi parallel inference as well as Anemoi-in-Cascade-non-parallel, in a mutually non disruptive way
- how to allow Anemoi parallel inference in Cascade, that is, slurm-within-slum :)

for the first one, it may be easiest to use custom env vars (ANEMOI_NODELIST, ANEMOI_JOBID, ...), and have the example bash submit script set them to slurm's vars. And for the second one... well, a more profound deliberation needs to be taken, but let's just keep it in mind laterally, and solve it later

src/anemoi/inference/runner.py
tmi: when solving the gpu device assignment problem in Anemoi for Cascade, for some reason I chose exporting CUDA_VISIBLE_DEVICES over torch.cuda.set_device, but can't remember exactly why... I think there was some weird case of model config default not playing nicely with this...

src/anemoi/inference/parallel.py
tmi:
```
Depending on build-time configurations, valid values include mpi, gloo, nccl, and ucc.
```
but if you mean the `tcp`, that's just for init afaik... And `nccl` is a good choice here

src/anemoi/inference/parallel.py
cathalobrien: I could add a `--num-gpus` flag to anemoi inference, which defaults to 1

src/anemoi/inference/parallel.py
cathalobrien: done, good catch

src/anemoi/inference/parallel.py
cathalobrien: Yeah to use MPI you have to build torch from source.

Gloo is intended for running on CPU. Actually It would be good to check device and pick between Gloo and NCCL accordingly

src/anemoi/inference/runner.py
cathalobrien: > We should be mindful of locking to cuda, but that is likely out of scope for this PR

Sure, i can wrap this in a `if device == cuda` block

src/anemoi/inference/runner.py
```diff
@@ -277,56 +293,67 @@ def forecast(self, lead_time, input_tensor_numpy, input_state):
...
             with torch.autocast(device_type=self.device, dtype=self.autocast):
-                y_pred = self.model.predict_step(input_tensor_torch)
+                if model_comm_group is None:
+                    y_pred = self.model.predict_step(input_tensor_torch)
+                else:
+                    try:
+                        y_pred = self.model.predict_step(input_tensor_torch, model_comm_group)
+                    except TypeError as err:
+                        LOG.error("Please upgrade to a newer version of anemoi-models to use parallel inference")
+                        raise err
```
gmertes, line 310: The call to `self.model.predict_step()` has changed in develop. We now call an encapsulation function `self.predict_step()` that deals with any optional arguments.

So can you move this code to be inside the new `self.predict_step()`.

gmertes (commented):
I'm a bit concerned about adding the parallel stuff to the default Runner. I would rather create a separate `ParallelRunner` for it. 

In develop we now have the option to instantiate runners from the config, so you would explicitly select it when you want to do parallel inference.

Opinions on factoring out this code into a new runner?

- Add a `Runner.log()` that wraps the logger. 
- Factor out the output block into a `Runner.output()`. 
- Then in the ParallelRunner overrides of those functions, check rank before calling `super()`
- Passing of the `model_comm_group` can be done in the same way as we did for the CRPS runner that's now in develop (see my other comment about predict_step)

jswijnands:
Thanks Cathal, this looks very interesting. I had a question about the slurm dependency to launch the parallel processes. Do you mean it will only work using slurm or could this also work on AWS?

cathalobrien:
> Thanks Cathal, this looks very interesting. I had a question about the slurm dependency to launch the parallel processes. Do you mean it will only work using slurm or could this also work on AWS?

Hi Jasper. Yeah currently you would need Slurm to get the 'srun' program.
For a single node case, we could fallback to launching subprocesses va python. For multi-node inference, we could use mpirun to launch the processes instead. 

Whether or not this works on AWS would depend on your exact setup. If you are using [AWS Parallel Cluster](https://docs.aws.amazon.com/parallelcluster/latest/ug/what-is-aws-parallelcluster.html) you could get slurm on that cluster. 

It would be nice to get more details about your setup to make sure it is supported. Could you send me an email (cathal.obrien@ecmwf.int) or message me on the Slack?

src/anemoi/inference/runner.py
cathalobrien: yeah I'd be happy to do that. Good idea

src/anemoi/inference/runner.py
```diff
@@ -277,56 +293,67 @@ def forecast(self, lead_time, input_tensor_numpy, input_state):
...
 
-            # Detach tensor and squeeze (should we detach here?)
-            output = np.squeeze(y_pred.cpu().numpy())  # shape: (values, variables)
+            if global_rank == 0:
```
ssmmnn11, line 312: pls see offline discussion :-)

src/anemoi/inference/runner.py
```diff
@@ -239,25 +241,39 @@ def model(self):
             return model
 
     def forecast(self, lead_time, input_tensor_numpy, input_state):
+
+        # determine processes rank for parallel inference and assign a device
+        global_rank, local_rank, world_size = get_parallel_info()
```
ssmmnn11, line 246: In training we seed the whole parallel model group with the same random [seed](https://github.com/ecmwf/anemoi-training/blob/c9fa231390cdb3da94bd93b48baaf82b5498f0ff/src/anemoi/training/distributed/strategy.py#L150) I think it would be nice to do the same here as well. The challenge is to generate appropriate seeds...

ssmmnn11 (changes requested):
nice work!

src/anemoi/inference/parallel.py
cathalobrien: You now have to add "runner: 'parallel'"to inference config to enable parallelisation. Currently, it still assumes you always want 'SLURM_PROCID' gpus, which might not suit all use cases

src/anemoi/inference/parallel.py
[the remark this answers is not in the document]
cathalobrien: @tmi I believe the first point is resolved now that parallelism is disabled by default.

src/anemoi/inference/runners/parallel.py
```diff
@@ -0,0 +1,146 @@
...
+        """Reads Slurm environment to set master address and port for parallel communication"""
+
+        # Get the master address from the SLURM_NODELIST environment variable
+        slurm_nodelist = os.environ.get("SLURM_NODELIST")
+        if not slurm_nodelist:
+            raise ValueError("SLURM_NODELIST environment variable is not set.")
+
+        # Use subprocess to execute scontrol and get the first hostname
+        result = subprocess.run(
+            ["scontrol", "show", "hostname", slurm_nodelist], stdout=subprocess.PIPE, text=True, check=True
```
cathalobrien, line 85: `scontrol` doesn't seem to work inside a container (exit code 127), Also it increases the dependency on slurm. So I will try find an alternative.

src/anemoi/inference/runner.py
```diff
@@ -244,18 +244,15 @@ def predict_step(self, model, input_tensor_torch, fcstep, **kwargs):
         return model.predict_step(input_tensor_torch)
 
     def forecast(self, lead_time, input_tensor_numpy, input_state):
-        self.model.eval()
```
gmertes, line 247: Why remove this?

src/anemoi/inference/runners/parallel.py
cathalobrien: Alternative now is to set optionally set MASTER_ADDR via an env var

docs/parallel.rst
```diff
@@ -1,54 +1,64 @@
-###################
+####################
  Parallel Inference
-###################
+####################
 
-If the memory requirements of your model are too large to fit within a single GPU, you run Anemoi-Inference in parallel across multiple GPUs.
+If the memory requirements of your model are too large to fit within a
+single GPU, you run Anemoi-Inference in parallel across multiple GPUs.
```
HCookie, line 6:
```suggestion
single GPU, you can run Anemoi-Inference in parallel across multiple GPUs.
```

docs/parallel.rst
```diff
@@ -1,54 +1,64 @@
...
 
-Parallel inference requires SLURM to launch the parallel processes and to determine information about your network environment. If SLURM is not available to you, please create an issue on the Anemoi-Inference github page.
+Parallel inference requires SLURM to launch the parallel processes and
+to determine information about your network environment. If SLURM is not
+available to you, please create an issue on the Anemoi-Inference github
+page.
```
HCookie, line 11:
```suggestion
page `here <https://github.com/ecmwf/anemoi-inference/issues>`_.
```

docs/parallel.rst
```diff
@@ -1,54 +1,64 @@
...
-        #SBATCH --cpus-per-task=8
-        #SBATCH --time=0:05:00
-        #SBATCH --output=outputs/paralell_inf.%j.out
+   #!/bin/bash
+   #SBATCH --nodes=1
+   #SBATCH --ntasks-per-node=4
+   #SBATCH --gpus-per-node=4
+   #SBATCH --cpus-per-task=8
+   #SBATCH --time=0:05:00
+   #SBATCH --output=outputs/paralell_inf.%j.out
```
HCookie, line 45:
```suggestion
   #SBATCH --output=outputs/parallel_inf.%j.out
```

HCookie (commented):
Looks really good,

docs/parallel.rst
cathalobrien: Done, thanks for the quick feedback!

gmertes (dismissed):
LGTM from my end, but others may still want to have a look.

src/anemoi/inference/runner.py
cathalobrien: resolved

src/anemoi/inference/parallel.py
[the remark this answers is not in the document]
cathalobrien: I will resolve this conversation now, issues with Anemoi parallel inference within cascade can be dealt with when they arise

ssmmnn11 (approved):

HCookie (approved):
Fantastic work on this PR. 
Looks **great** to me

PR #109, opened 2025-01-19
feat: Add support for `output_frequency` to write less output

b8raoult:
Allow users to write only certain steps. The frequency can be set globally or per output.

### DIFF
--- a/src/anemoi/inference/output.py
+++ b/src/anemoi/inference/output.py
@@ -9,6 +9,7 @@
 import logging
 from abc import ABC
 from abc import abstractmethod
+from functools import cached_property
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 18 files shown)


src/anemoi/inference/output.py
```diff
@@ -6,26 +6,67 @@
...
+    def write_initial_step(self, state, step):
+        """This method should not be called directly
+        call `write_initial_state` instead.
+        """
+        reduced_state = self.reduce(state)
+        self.write_step(reduced_state, step)
+
+    @abstractmethod
+    def write_step(self, state, step):
+        """This method should be be called directly
```
dietervdb-meteo, line 67:
```suggestion
        """This method should not be called directly
```

dietervdb-meteo:
Should it be enforced/checked if `output_frequency` is a multiple of the model timestep? In the current implementation once can set up a 24h lead time inference, put `output_frequency=4` or `output_frequency=5` for a 6h step model and inference will run without problems. In the first case the actual `output_frequency` will be 12, the lcm, while in the second case no output will be generated since the lcm, 30, is outside of the lead time.

floriankrb (approved):

PR #111, opened 2025-01-21
feat: Use array_api when interfacing with data

HCookie:
# Flexible Data Types

Allow for predictions to stay on device
- Use array_api to interface with data when needed


HCookie:
If https://github.com/ecmwf/earthkit-data/pull/598 is accepted, this can be simplified

PR #118, opened 2025-01-30
feat(output): allow selecting variables for GRIB output

frazane:
# Purpose
Allow users to specify a list of variables (a subset of the variables in the state fields) that they want to include in an output GRIB file. 

# Code changes
- added a `variable` keyword to the corresponding classes. When not provided, all fields are written to output. 
- also made a bugfix on L262 of `grib.py` so it uses the correct param (based on the original local definitions used when creating the anemoi dataset), e.g. in our case `T_2M` would need to be used instead of `2t`.

### DIFF
--- a/src/anemoi/inference/outputs/grib.py
+++ b/src/anemoi/inference/outputs/grib.py
@@ -83,6 +83,7 @@ class GribOutput(Output):
         modifiers=None,
         output_frequency=None,
         write_initial_state=None,
+        variables=None,
     ):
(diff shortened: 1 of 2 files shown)


frazane:
I didn't find any pre-existing tests for the output filename formatting so I didn't add any myself to test this new feature, because I am not sure about the best way to do it. Suggestions are welcome!

HCookie (approved):
Thanks for the contribution.

PR #121, opened 2025-02-03
feat: parallel inference without slurm

cathalobrien:
This PR extends the Parallel Inference added in #55 to work without slurm **within 1 node**. It's much nicer to debug and run now :D

When running parallel inference without slurm, you have to add `world_size` to the config. At the moment, this is ignored in favour of `SLURM_NTASKS` when running with srun. An example of a config running parallel inference across 4 nodes is shown below. You can launch this job as normal with `anemoi-inference run parinf.yaml`.

```yaml
checkpoint: /path/to/inference-last.ckpt
lead_time: 60
runner: parallel
world_size: 4 #Only required if running parallel inference without Slurm
input:
  grib: /path/to/input.grib
output:
  grib: /path/to/output.grib
```

How it works is:
- check if anemoi-inference is launched by srun
- if not, spawn `config.world_size` processes
   - `master_addr` is `localhost`
   - `master_port` is a hash of the node name, within a range. 
- Each spawned process runs a slimmed down version `RunCmd.run` but with the config preloaded

Issues:

- The master port calculation would lead to a clash if two parallel inference processes ran on the same node at the same time. 
- At the moment, I have a copy of `RunCmd.run` in `runners/parallel.py`. Would be nice to be able to use that code directly, rather then having to maintain a copy. To do this, I would just have to be able to pass a loaded config instead of a path

----
📚 Documentation preview 📚: https://anemoi-inference--121.org.readthedocs.build/en/121/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,6 @@ Keep it human-readable, your future self will thank you!
 - Fix `__version__` import in init
 - use earthkit-data 0.11.2
 - Fix SimpleRunner
-- Inference can be run in parallel on 1 node without slurm [#121](https://github.com/ecmwf/anemoi-inference/pull/121)
 
 ### Removed
 - ci: turn off hpc workflow
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Keep it human-readable, your future self will thank you!
 - Fix `__version__` import in init
 - use earthkit-data 0.11.2
 - Fix SimpleRunner
+- Inference can be run in parallel on 1 node without slurm [#121](https://github.com/ecmwf/anemoi-inference/pull/121)
 
 ### Removed
 - ci: turn off hpc workflow
--- a/src/anemoi/inference/config.py
+++ b/src/anemoi/inference/config.py
@@ -53,6 +53,12 @@ class Configuration(BaseModel):
     verbosity: int = 0
     """The verbosity level of the runner. This can be 0 (default), 1, 2 or 3."""
 
+    pid: Optional[int] = 0
+    """ Process id, used for parallel inference without SLURM """
+
+    world_size: Optional[int] = 1
+    """ Number of parallel processes, used for parallel inference without SLURM """
+
     report_error: bool = False
     """If True, the runner list the training versions of the packages in case of error."""
 
--- a/src/anemoi/inference/config.py
+++ b/src/anemoi/inference/config.py
@@ -53,9 +53,6 @@ class Configuration(BaseModel):
     verbosity: int = 0
     """The verbosity level of the runner. This can be 0 (default), 1, 2 or 3."""
 
-    pid: Optional[int] = 0
-    """ Process id, used for parallel inference without SLURM. This should not be set by the user"""
-
     world_size: Optional[int] = 1
     """ Number of parallel processes, used for parallel inference without SLURM """
 
--- a/src/anemoi/inference/config.py
+++ b/src/anemoi/inference/config.py
@@ -54,7 +54,7 @@ class Configuration(BaseModel):
     """The verbosity level of the runner. This can be 0 (default), 1, 2 or 3."""
 
     pid: Optional[int] = 0
-    """ Process id, used for parallel inference without SLURM """
+    """ Process id, used for parallel inference without SLURM. This should not be set by the user"""
 
     world_size: Optional[int] = 1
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ Keep it human-readable, your future self will thank you!
 - Fix `__version__` import in init
 - use earthkit-data 0.11.2
 - Fix SimpleRunner
+- Inference can be run in parallel on 1 node without slurm [#121](https://github.com/ecmwf/anemoi-inference/pull/121)
```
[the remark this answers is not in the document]
cathalobrien: ok. removed

cathalobrien:
note to self: update parallel docs to clearly state it requires a minimum version of anemoi-models

src/anemoi/inference/runners/parallel.py
```diff
@@ -88,7 +81,9 @@ def predict_step(self, model, input_tensor_torch, fcstep, **kwargs):
             try:
                 return model.predict_step(input_tensor_torch, self.model_comm_group)
             except TypeError as err:
-                LOG.error("Please upgrade to a newer version of anemoi-models to use parallel inference")
+                LOG.error(
+                    "Please upgrade to a newer version of anemoi-models (at least commit 'db587fe5b350a98cfb95b90ede327e6ec6d470da') to use parallel inference. If updating breaks your checkpoints, you can try reverting to your original version of anemoi-models and cherry-picking 'https://github.com/ecmwf/anemoi-core/pull/77'"
+                )
```
HCookie, line 86: Is a commit tag necessary, do we have a minimum version to support?

src/anemoi/inference/runners/parallel.py
```diff
@@ -104,14 +99,41 @@ def __del__(self):
...
+
+        seed = None
+        seed_threshold = 1000
+        env_var_list = ["ANEMOI_BASE_SEED"]
+        for env_var in env_var_list:
+            if env_var in os.environ:
+                seed = int(os.environ.get(env_var))
+                if seed < seed_threshold:
+                    seed *= seed_threshold  # make it (hopefully) big enough
+                break
```
HCookie, line 115: Why is there a seed threshold?

src/anemoi/inference/runners/parallel.py
cathalobrien: Not currently, the minimum version would be the next model release

src/anemoi/inference/runners/parallel.py
cathalobrien: No idea, It's also present in anemoi-training and I wanted to keep the seed function the same

src/anemoi/inference/runners/parallel.py
```diff
@@ -19,45 +19,67 @@
...
+    input_state = input.create_input_state(date=config.date)
+
+    output.write_initial_state(input_state)
+
+    for state in runner.run(input_state=input_state, lead_time=config.lead_time):
+        output.write_state(state)
+
+    output.close()
+
+
```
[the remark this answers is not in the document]
gmertes: @cathalobrien Could you abstract this into a function that is shared between the RunCmd and here? In `commands/run.py` add a `_run(config)` function at the top level with these lines, and then use that one in both places? 

Because there are already upcoming changes to this logic in the RunCmd, that won't be maintainable otherwise.

src/anemoi/inference/config.py
```diff
@@ -53,6 +53,12 @@ class Config:
     verbosity: int = 0
     """The verbosity level of the runner. This can be 0 (default), 1, 2 or 3."""
 
+    pid: Optional[int] = 0
+    """ Process id, used for parallel inference without SLURM. This should not be set by the user"""
```
gmertes, line 57: Does this need to be in the config if it's not supposed to be set by the user? Can it just be a member of the runner class?

src/anemoi/inference/runners/parallel.py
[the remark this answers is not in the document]
cathalobrien: great idea

src/anemoi/inference/runners/parallel.py
cathalobrien: 4.2 is out now so will change

src/anemoi/inference/config.py
cathalobrien: deleted in latest commit

gmertes (approved):

PR #144, opened 2025-02-18
feat: Add truth output

HCookie:
## Description

Add a truth output, which writes the truth from the input at the same time.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

-   [x] I have tested that new dependencies themselves are pip-installable.

### Documentation

-   [x] I have updated the documentation and docstrings to reflect the changes
-   [x] I have added comments to my code, particularly in hard-to-understand areas

## Additional Notes

----
📚 Documentation preview 📚: https://anemoi-inference--144.org.readthedocs.build/en/144/

### DIFF
--- a/src/anemoi/inference/outputs/truth.py
+++ b/src/anemoi/inference/outputs/truth.py
@@ -10,6 +10,7 @@
 import logging
 
 from ..context import Context
+from ..output import ForwardOutput
 from ..output import Output
 from . import create_output
 from . import output_registry
@@ -18,7 +19,7 @@ LOG = logging.getLogger(__name__)
 
 
 @output_registry.register("truth")
-class TruthOutput(Output):
+class TruthOutput(ForwardOutput):
     """Write the truth state for each output state."""
 
     def __init__(self, context: Context, output: dict, output_frequency=None, write_initial_state=None, **kwargs):
--- a/src/anemoi/inference/outputs/truth.py
+++ b/src/anemoi/inference/outputs/truth.py
@@ -22,8 +22,8 @@ LOG = logging.getLogger(__name__)
 class TruthOutput(ForwardOutput):
     """Write the truth state for each output state."""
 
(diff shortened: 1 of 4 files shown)


src/anemoi/inference/outputs/truth.py
```diff
@@ -0,0 +1,52 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+@output_registry.register("truth")
+class TruthOutput(ForwardOutput):
+    """Write the truth state for each output state."""
+
+    def __init__(self, context: Context, output: dict, output_frequency=None, write_initial_state=None, **kwargs):
+        super().__init__(context, output_frequency=output_frequency, write_initial_state=write_initial_state)
```
gmertes, line 26: Since you are not using or modifying the arguments of the super call, I would just catch all named arguments with **kwargs and pass those through to super.init. So that it is forward compatible with any changes to the super class constructor. 
But that's a personal preference, you can disagree if you want the signature to be explicit.

src/anemoi/inference/outputs/truth.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.output: Output = create_output(context, output)
+
+    def write_initial_state(self, state):
+        self.output.write_initial_state(state)
+
+    def write_state(self, state):
+        truth_state = self.context.create_input().create_input_state(date=state["date"])
```
gmertes, line 34: Creating an input could potentially do work, so I would create the input once and reuse it. Unless you see a possibility where the created input could depend on the state, but right now that shouldn't be the case?

src/anemoi/inference/outputs/truth.py
HCookie: Yep, I agree, was copied from the tee

src/anemoi/inference/outputs/truth.py
HCookie: Good point, will move to init

gmertes (approved):

PR #169, opened 2025-03-06
feat: support for cutout input

frazane:
## Description

Adds support to run inference for the "stretched-grid" approach by introducing a new `cutout` input type. This new class specifies the logic to combine one or more LAM source(s) nested in a global source, using the supporting arrays stored in a checkpoint.

Example config:
```yaml
input:
  cutout:
    lam_0:
      grib:
        path: <path_to_grib>
    global:
      grib: 
        path: <path_to_grib>
```

## Type of Change

-   [x] Documentation update

## Code Compatibility

----
📚 Documentation preview 📚: https://anemoi-inference--169.org.readthedocs.build/en/169/


src/anemoi/inference/inputs/gribfile.py
```diff
@@ -27,11 +27,12 @@ class GribFileInput(GribInput):
     trace_name = "grib file"
 
     def __init__(self, context, path, *, namer=None, **kwargs):
-        super().__init__(context, namer=namer, **kwargs)
+        super().__init__(context, namer=namer)
         self.path = path
+        self.kwargs = kwargs
```
gmertes, line 32: Using the kwargs like this is dangerous. For now this works because the super class doesn't use them, but if in the future the super class starts accepting any kwargs it will break.

Why don't we make `mask` part of the constructor of `EkdInput`, as an optional parameter? You then store it in a `self.mask` and pass it through where needed. It should be compatible with all other ekd-derived inputs.

src/anemoi/inference/inputs/ekd.py
```diff
@@ -133,15 +117,24 @@ def _create_input_state(
         fields = input_state["fields"]
 
         input_fields = self._filter_and_sort(input_fields, variables=variables, dates=dates, title="Create input state")
-        mask = self.checkpoint.grid_points_mask
-        mask = ApplyMask(mask) if mask is not None else NoMask()
 
+        # target information
+        if mask is not None:
+            LOG.info("Applying mask from supporting arrays: %s", mask)
+            mask = self.checkpoint.load_supporting_array(mask)
```
gmertes, line 124: Could you raise a meaningful error here if this fails? Something like `"An input mask {mask} was supplied but the supporting array was not found in the checkpoint"` or something like that.

src/anemoi/inference/inputs/cutout.py
```diff
@@ -0,0 +1,72 @@
...
+    def create_input_state(self, *, date=None):
+        """Create the input state dictionary."""
+
+        LOG.info(f"Concatenating states from {self.sources}")
+        src = list(self.sources.keys())
+        state = self.sources[src[0]].create_input_state(date=date)
+        for i, source in enumerate(src[1:]):
+            _state = self.sources[source].create_input_state(date=date)
+
+            # NOTE: we must decide whether these come from the checkpoint or the input
```
gmertes, line 54: Do we need to decide that here? The underlying ekdinput should take care of that - using lat/lon from the checkpoint if they exist, otherwise try to use them from the input.

src/anemoi/inference/inputs/ekd.py
```diff
@@ -150,15 +143,15 @@ def _create_input_state(
...
+                    shape=(len(dates), n_points),
                     fill_value=np.nan,
                     dtype=dtype,
                 )
 
             date_idx = date_to_index[valid_datetime]
 
             try:
-                fields[name][date_idx] = mask.apply(field.to_numpy(dtype=dtype, flatten=flatten))
+                fields[name][date_idx] = field.to_numpy(dtype=dtype, flatten=flatten)[mask]
```
gmertes, line 154: In the case of no mask being supplied, this will apply a mask equal to the number of grid points (or a slice of None). Does this introduce an extra operation, or is numpy smart enough to treat that as a noop?

Before with the mask class and `apply()` we were sure that not having a mask would be a noop. We should be sure that that is still the case.

gmertes (commented):
Thanks for the contribution! It looks good, just some minor comments.

src/anemoi/inference/inputs/ekd.py
frazane: There are already meaningful messages further down: https://github.com/ecmwf/anemoi-inference/blob/cc7fed54aa5f1923da7ef08649b102e2e690a3bf/src/anemoi/inference/metadata.py#L729 

is this enough or should I add something here as well?

src/anemoi/inference/inputs/cutout.py
frazane: Yes, sorry, that's what I meant. I just wasn't sure about touching that part, we might leave it for another PR but I left a note. For the time being the checkpoint is still used.

src/anemoi/inference/inputs/ekd.py
frazane: If no mask is provided, mask is `slice(None)` which I would assume is interpreted as a noop by numpy.  But yeah, I am not 100% sure. This small test would suggest so:

```python
import numpy as np
x = np.arange(10)
y = x[range(10)]
print(np.shares_memory(x, y)) # False
print(np.all(x == y)) # True
y = x[slice(None)]
print(np.shares_memory(x, y)) # True
print(np.all(x == y)) # True
```

src/anemoi/inference/inputs/gribfile.py
frazane: I saw that extra kwargs were passed to `_create_input_state` in the `EkdInput` instead of to the constructor so I  did the same. I agree it's not ideal. I'll edit this, thanks!

src/anemoi/inference/inputs/ekd.py
gmertes: Ah yes, that is enough.

gmertes (commented):
Thanks for the changes! I'm happy with this, but I also want to do some testing myself to ensure backwards compatibility, and also give colleagues a chance to review. So I'm going to wait until after the weekend to approve this, if that's ok.

JPXKQX:
Thanks for the PR Francesco, this functionality will be really useful for other users in Anemoi training stretched graphs. I think that adding some information about this in the docs would be very helpful for them.

src/anemoi/inference/inputs/cutout.py
```diff
@@ -0,0 +1,72 @@
...
+            # state["latitudes"] = np.concatenate(
+            #     [state["latitudes"], _state["latitudes"]], axis=-1
+            # )
+            # state["longitudes"] = np.concatenate(
+            #     [state["longitudes"], _state["longitudes"]], axis=-1
+            # )
+            for field, values in state["fields"].items():
+                state["fields"][field] = np.concatenate([values, _state["fields"][field]], axis=-1)
```
gmertes, line 62: Sorry to be a bother Francesco, but after discussing with Baudouin we think it might be better to move the masking completely to the Cutout input. So the opposite of what I suggested before (sorry!).

We make `mask` a parameter of this input (so in the config, it moves up 1 level), you get the original fields from each source with `source.create_input_state`, and you apply the mask on that and concatenate. You can strip out all the old masking code out of the ekdinput, and then it is nicely separated. 

What do you think?

frazane: I am not sure I understand. The mask is different for each source - you cannot apply the same mask. 

But anyway, thinking about it, in theory using the `mask` argument is not _really_ needed since it's implicit that we will be using the cutout masks from the supporting arrays of the checkpoint.

gmertes: You would still have a mask per source, like this:

```yaml
input:
  cutout:
    lam_0:
      mask: lam_0/cutout_mask
      grib:
        path: <path_to_grib>
    global:
      mask: global/cutout_mask
      grib: 
        path: <path_to_grib>
```

Then in your Cutout input you maintain a list of masks as well as a list of sources. In the constructor, something like:

```python
self.sources = {}
self.masks = {}
for src, cfg in sources.items():
    self.masks[src] = cfg.pop("mask")
    self.sources[src] = create_input(context, cfg)
```

And then you apply the mask to each source when creating the input state.

> in theory using the mask argument is not really needed

Yes, I agree, but we could still have `mask` as an optional config entry with a default value of `{source_name}/cutout_mask`

frazane:
@JPXKQX I have added some documentation, feedback is welcome

gmertes (commented):
LGTM, proposing it to ATS this week.

gmertes (approved):

PR #182, opened 2025-03-19
chore: refactor inputs

frazane:
## Description

Refactoring of the inputs modules. The goal is to significantly reduce the duplicate code and to improve clarity and consistency.  No breaking changes.

I tried to limit changes to the rest of the codebase as much as possible, as I am planning to propose other refactorings gradually. Nevertheless, some small modifications had to be done to forcings and runners. 

### Summary of changes
- general refactoring of the individual methods on the input classes
- reorganization of the class hierarchies
- changed handling of single vs multiple dates for loading inputs, which is now consistent across different methods. Now, we always pass a single `date` along with a `initial: bool` keyword argument to indicate whether we should load the lagged past inputs or not.

### Tested inputs
- `dataset`✔
- `grib` ✔
- `cutout` ✔
- `dummy` ✔
- `icon_grib_file` ❌
- `mars` ❌
- `cds` ❌

## Type of Change

-   [ ] Documentation update

## Code Compatibility

-   [ ] I have performed a self-review of my code

### Dependencies

### Documentation


frazane:
@gmertes @b8raoult I marked this PR as ready for review because I'd like to get some early feedback, when you have some time. But there's still some missing components that need to be tested (see list above).

gmertes:
Hi @frazane sorry for the long wait, I will find some time in the next few days to have a look at this.

src/anemoi/inference/forcings.py
```diff
@@ -132,29 +134,14 @@ def __repr__(self) -> str:
...
 
-        if not isinstance(dates, (list, tuple)):
-            dates = [dates]
-
+        dates = [date + h for h in self.checkpoint.lagged] if initial else [date]
         source = UnstructuredGridFieldList.from_values(
-            latitudes=current_state["latitudes"],
-            longitudes=current_state["longitudes"],
+            latitudes=self.checkpoint.latitudes,
+            longitudes=self.checkpoint.longitudes,
```
gmertes, line 144: Having this hardcoded here is a problem. We need the logic of first trying to take the lat/lon from the input fields, and then using the checkpoint as a fallback. That seems to have disappeared.

src/anemoi/inference/forcings.py
```diff
@@ -263,29 +237,16 @@ def __repr__(self) -> str:
...
-            self.input.load_forcings_state(variables=self.variables, dates=dates, current_state=current_state),
+            self.input.load_forcings_state(date=date, variables=self.variables, initial=initial),
             self.variables,
-            dates,
+            [date],
         )
         data = data[..., self.spatial_mask]
 
-        expected_shape = (len(self.variables), len(dates), current_state["latitudes"][self.spatial_mask].size)
+        expected_shape = (len(self.variables), 1, self.checkpoint.latitudes[self.spatial_mask].size)
```
gmertes, line 249: Same issue here.

src/anemoi/inference/inputs/ekd.py
```diff
@@ -128,298 +106,61 @@ def __init__(
         self._namer = namer if namer is not None else self.checkpoint.default_namer()
         assert callable(self._namer), type(self._namer)
 
-    def _filter_and_sort(self, data: Any, *, variables: List[str], dates: List[Any], title: str) -> Any:
-        """Filter and sort the data.
+    def create_state(self, *, date=None, variables=None, initial: bool = True) -> State:
```
gmertes, line 109: Why did you choose the default `initial` as True? Would it not make more sense the other way around?

src/anemoi/inference/inputs/ekd.py
```diff
@@ -128,298 +106,61 @@ def __init__(
...
-        data : Any
-            The data to filter and sort.
-        variables : List[str]
-            The list of variables to select.
-        dates : List[Any]
-            The list of dates to select.
-        title : str
-            The title for logging.
+        # define request (using checkpoint)
+        variables = self.checkpoint_variables if variables is None else variables
```
gmertes, line 113: Using `self.checkpoint_variables` here is wrong. That one contains all input variables including the computed forcings. The state here should be created only from non-computed variables, since they are added to it afterwards.

You need to use:
`self.checkpoint.variables_from_input(include_forcings=True)`
This one skips computed forcings.

src/anemoi/inference/inputs/ekd.py
```diff
@@ -128,298 +106,61 @@ def __init__(
...
+        date = np.datetime64(date).astype(datetime.datetime)
```
gmertes, line 114: The logic of what happens when `date` is None has disappeared, I'm unable to run without specifying a date now.

When running from a grib file or dataset, it should pick the latest available date in the fieldlist. That logic used to be in the ekd.py  `_create_input_state`. When running from mars or cds, it should take yesterday's date.

You should also try to keep support for specifying the date like `-1` , which used to be handled by the `to_datetime()` utility.

src/anemoi/inference/inputs/mars.py
```diff
@@ -219,43 +216,15 @@ def __init__(
...
-            self.retrieve(
-                self.variables,
-                [date + h for h in self.checkpoint.lagged],
-            ),
-            variables=self.variables,
-            date=date,
-        )
-
-    def retrieve(self, variables: List[str], dates: List[Date]) -> Any:
+    def _earthkit_reader(self, variables: List[str], dates: List[Date]) -> Any:
```
gmertes, line 227: The order of the arguments is wrong, dates and variables should be swapped.

radiradev:
Now that integration tests are in place (https://github.com/ecmwf/anemoi-inference/issues/271) should this be re-opened?

PR #184, opened 2025-03-20
fix: Add check for date is None with earthkit-data GRIB input

gareth-j:
## Description

Adds a check if the date is None with the earthkit-data GRIB input.

## Type of Change

-   [ ] Documentation update

### DIFF
--- a/src/anemoi/inference/inputs/ekd.py
+++ b/src/anemoi/inference/inputs/ekd.py
@@ -413,7 +413,14 @@ class EkdInput(Input):
         State
             The created input state.
         """
+        if date is None:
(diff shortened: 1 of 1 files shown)


src/anemoi/inference/inputs/ekd.py
```diff
@@ -413,7 +413,14 @@ def _create_input_state(
         State
             The created input state.
         """
+        if date is None:
```
frazane, line 416: If `date` can be `None`, the type hint should also be changed to `Optional[Date]`.

gareth-j: Yes! Fixed.

gmertes (approved):

PR #186, opened 2025-03-21
fix: Write correct values at step 0

gmertes:
## Description

When writing the initial state the multistep input, a 2D array containing t0 and t-6, is written to step 0. Only the 1D array of t0 should be written to step 0.

Thanks for @frazane for finding the bug and proposing the solution.

## Type of Change

-   [ ] Documentation update


HCookie:
Why not use the `reduce` method?
https://github.com/ecmwf/anemoi-inference/blob/c645fb219ed40ea60389affafe0f4f6c3b799ee9/src/anemoi/inference/output.py#L91-L109

gmertes:
I forgor 💀

Using reduce now.

PR #190, opened 2025-03-24
feat(input): add FDB input class

frazane:
## Description

This PR introduces a new class to get input fields from [FDB](https://github.com/ecmwf/fdb). For using it, one must specify some keys of the request as kwargs in the configuration, which are then completed/overridden by the keys (date, time, param, etc.) of the mars request for the data as specified in the checkpoint (just like in `MarsInput`). The FDB configuration can either be provided as additional key-value pairs, or as environment variables.

### Example config
```yaml
input:
    fdb:
      expver: 0001
      number: 0
      step: 0
      class: od
      levtype: sfc
      model: ICON-CH1-EPS
      stream: enfo
      type: ememb
```

## Type of Change

-   [ ] Documentation update

## Code Compatibility

-   [ ] I have performed a self-review of my code

### Dependencies

### Documentation


b8raoult (commented):
You do not need the "request_template" part. This will be more inline with the other inputs and datasets sources

frazane:
> You do not need the "request_template" part. This will be more inline with the other inputs and datasets sources

Hi @b8raoult, my understanding is that `fdb` requests must be fully specified (is that correct?) However, we do not get a fully specified request from the checkpoint alone, and some missing keys need to be completed somehow. A request template seemed like the most straightforward approach but if there is a better way we should discuss it.

frazane:
Remove the `request_template` - now additional keys are passed as kwargs.

src/anemoi/inference/inputs/fdb.py
```diff
@@ -0,0 +1,90 @@
...
+        fdb_userconfig : dict, optional
+            The FDB userconfig to use.
+        kwargs : dict, optional
+            Additional keyword arguments for the request to FDB.
+        """
+        super().__init__(context, namer=namer)
+        self.kwargs = kwargs
+        self.configs = {"config": fdb_config, "userconfig": fdb_userconfig}
+        self.fdb_config = fdb_config
+        self.fdb_userconfig = fdb_userconfig
```
HCookie, line 61: These don't appear to be used

HCookie (approved):
Thanks, looks good.
@b8raoult any comments?

frazane:
Yes. Thanks for the support!

PR #194, opened 2025-03-25
feat: Add version validation if model fails to load

HCookie:
## Description

Add environment context information if the model fails to load.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/anemoi/inference/runner.py
+++ b/src/anemoi/inference/runner.py
@@ -452,7 +452,11 @@ class Runner(Context):
             The loaded model.
         """
         with Timer(f"Loading {self.checkpoint}"):
-            model = torch.load(self.checkpoint.path, map_location=self.device, weights_only=False).to(self.device)
+            try:
+                model = torch.load(self.checkpoint.path, map_location=self.device, weights_only=False).to(self.device)
+            except Exception as e:
+                self.checkpoint.validate_environment(on_difference="warn")
+                raise RuntimeError("Error loading model, check the logs to see if the environment is valid.") from e
             # model.set_inference_options(**self.inference_options)
             assert getattr(model, "runner", None) is None, model.runner
             model.runner = self


src/anemoi/inference/runner.py
```diff
@@ -452,7 +452,11 @@ def model(self) -> torch.nn.Module:
             The loaded model.
         """
         with Timer(f"Loading {self.checkpoint}"):
-            model = torch.load(self.checkpoint.path, map_location=self.device, weights_only=False).to(self.device)
+            try:
+                model = torch.load(self.checkpoint.path, map_location=self.device, weights_only=False).to(self.device)
+            except (AttributeError, ValueError) as e:
```
aaron-hopkinson, line 457: Copying over from Teams for visibility...

If `anemoi-models` isn't installed (I think we need to add that to the `pyproject.toml`) we'll get a `ModuleNotFoundError`, I suspect a version mismatch might lead to a `pickle.UnpicklingError` as well (but I haven't checked). That said, I haven't checked to see if `torch.load` does any special exception handling itself...

src/anemoi/inference/runner.py
```diff
@@ -452,7 +452,11 @@ def model(self) -> torch.nn.Module:
...
+                self.checkpoint.validate_environment(on_difference="warn")
+                raise RuntimeError("Error loading model, check the logs to see if the environment is valid.") from e
```
aaron-hopkinson, line 459: `validate_environment` can raise a `RuntimeError` itself (`on_difference="error"`), is there a reason not to just use that rather than raising yourself?

aaron-hopkinson (changes requested):

src/anemoi/inference/runner.py
HCookie: Question is how to raise it as a sub error of this one?

aaron-hopkinson: Ah, gotcha - just relying on `validate_environment` to raise won't be clear why/where the original error came from. I think that's a good argument for leaving it as it is :)

src/anemoi/inference/runner.py
HCookie: I've addressed this in the latest commit, can you take a look?

src/anemoi/inference/provenance.py
```diff
@@ -41,13 +45,33 @@
...
+) -> str: ...
+
+
+def validate_environment(
+    metadata: "Metadata",
+    *,
+    all_packages: bool = False,
+    on_difference: Literal["warn", "error", "ignore", "return"] = "warn",
+    exempt_packages: Optional[List[str]] = None,
+) -> Union[bool, str]:
```
aaron-hopkinson, line 74: To be honest, I'm not sure I like all of the different ways of returning/exiting from this function – the fact that we now have a set of function prototypes here is kind of indicative (to me) of the fact that we should probably be splitting this into multiple functions...  It looks like it does what it's supposed to, but feels like it needs a refactor.

aaron-hopkinson (commented):
Some minor comments - may be out of the scope of this PR, but I feel like we might want to clean this code at some point.

Also if we're adding a new feature, we should probably add a test - just something simple like mock/patching the call to `torch.load`, forcing it to raise an exception, and ensuring that the `validate_environment` is called is probably sufficient to check that the high-level behaviour is doing what we expect (rather than the details of what's returned etc).

src/anemoi/inference/runner.py
```diff
@@ -452,7 +452,11 @@ def model(self) -> torch.nn.Module:
             The loaded model.
         """
         with Timer(f"Loading {self.checkpoint}"):
-            model = torch.load(self.checkpoint.path, map_location=self.device, weights_only=False).to(self.device)
+            try:
+                model = torch.load(self.checkpoint.path, map_location=self.device, weights_only=False).to(self.device)
+            except Exception as e:  # Wildcard exception to catch all errors
+                error_msg = f"Error loading model - {self.checkpoint.validate_environment(on_difference='return')}"
```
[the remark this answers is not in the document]
HCookie: Fixed in next commit

src/anemoi/inference/provenance.py
```diff
@@ -139,7 +167,7 @@ def validate_environment(
 
     for git_record in train_environment["git_versions"].keys():
         file_record = train_environment["git_versions"][git_record]["git"]
-        if file_record["modified_files"] == 0 and file_record["untracked_files"] == 0:
+        if file_record["modified_files"] == 0 and file_record["untracked_files"] == 0 or git_record in exempt_packages:
```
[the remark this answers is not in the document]
HCookie: Yes, it was. 
Fixed in next commit

src/anemoi/inference/checkpoint.py
```diff
@@ -258,24 +258,25 @@ def validate_environment(
...
 
         Parameters
         ----------
         all_packages : bool, optional
-            Whether to validate all packages, by default False.
-        on_difference : str, optional
-            Action to take on difference, by default "warn".
-        exempt_packages : Optional[List[str]], optional
-            List of packages to exempt, by default None.
+            Check all packages in environment or just `anemoi`'s, by default False
```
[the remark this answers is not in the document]
HCookie: Fixed in next commit

src/anemoi/inference/provenance.py
HCookie: I'm not sure, given how nested the call gets, from the runner passing to the checkpoint, the checkpoint then to the metadata, splitting into two functions may require a function on each level, which seems over complicated.

src/anemoi/inference/runner.py
aaron-hopkinson: On balance I think catching all exceptions might be the best, even though it feels a bit messy - marking as resolved. Thanks

src/anemoi/inference/provenance.py
aaron-hopkinson: Maybe just something to keep our eye on if it gets more complex in the future?

gmertes (approved):
LGTM. I share Aaron's concerns but we also want this function to be flexible. We could look at improving the documentation a bit, but it's good enough imo.

aaron-hopkinson (approved):

Issue #199, opened 2025-04-03
pydantic ConfigDict prevents passing new config entries via the command line override


mpvginde:
Hi,

I have the following use-case:  
I want to run LAM-inference for different dates, where all the initial states are stored in one `zarr`-file and the boundary conditions are stored in a `zarr`-file per forecast-run. 

To prevent me from rewriting the config for every inference run, I have the following config setup:
```yaml
checkpoint: 763af34c15134d1da1916f15ec502ada/inference-anemoi-by_epoch-epoch_011-step_150456.ckpt

input:
  test:
    use_original_paths: false

date: 2020-01-02T00:00:00
lead_time: 60h

input:
  dataset:
    cutout:
      - dataset: /datasets/ifs/ic/ifs-analysis_2024_v2.zarr

 output:
  extract_lam:
    lam: '0/lam_0'
    output:
      netcdf:
        path ./inference/2020010200.nc
```

The actual forecast date and boundary dataset are set by using the overrides e.g.
```bash
anemoi-inference run ./config/lam-512-600km-ifs.yaml \
  "date=2024-01-22T00:00:00"  \
  "+input.dataset.cutout={dataset: /datasets/ifs/lbc/ifs-lbc_20240122_00_v2.zarr}"  \
  "output.extract_lam.output.netcdf.path=./output/ifs-coupled/ifs-coupled_20240122_00.nc"
```

When I try this `pydantic` complains with
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Configuration
+input
  Extra inputs are not permitted [type=extra_forbidden, input_value={'dataset': {'cutout': '{..._20240122_00_v2.zarr}'}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden
2025-04-03 14:43:13 ERROR 
💣 1 validation error for Configuration
+input
  Extra inputs are not permitted [type=extra_forbidden, input_value={'dataset': {'cutout': '{..._20240122_00_v2.zarr}'}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/extra_forbidden
2025-04-03 14:43:13 ERROR 💣 Exiting
```

I guess this is because the `model_config` set here with keyword `extra="forbid"`
https://github.com/ecmwf/anemoi-inference/blob/a082af43ee214ad956315ef357f5d8d68842a662/src/anemoi/inference/config/__init__.py#L32-L36

[...]

PR #201, opened 2025-04-04
feat: allow hydra-like override of config list elements

mpvginde:
## Description

This PR allows `hydra`-like command line overrides of config list elements.

With `config.yaml`:
```
checkpoint: my_checkpoint.ckpt

input:
  test:
    use_original_paths: false

date: 2024-01-02T00:00:00
lead_time: 60h

input:
  dataset:
    cutout:
      - dataset: lam_analysis.zarr
      - dataset: ifs_lbc_20240101_00.zarr
    min_distance_km: 0
    adjust: all
output:
  extract_lam:
    lam: '0/lam_0'
    output:
      netcdf:
        path: my_forecast_20240102_00.nc
```

The second element in the list `input.dataset.cutout` can be overridden on the command line with:
`anemoi-inference run config.yaml "input.dataset.cutout.1.dataset=ifs_lbc_20240102_00.zarr"`

## Type of Change

-   [ ] Documentation update

## Issue Number

Alternative solution to #199 

## Code Compatibility

### Dependencies

### Documentation

----
📚 Documentation preview 📚: https://anemoi-inference--201.org.readthedocs.build/en/201/


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-inference?pullRequest=201) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-inference?pullRequest=201) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-inference?pullRequest=201) it.</sub>

src/anemoi/inference/config/__init__.py
```diff
@@ -89,7 +89,12 @@ def load(
                 key, value = override.split("=")
                 keys = key.split(".")
                 for key in keys[:-1]:
-                    path = path.setdefault(key, {})
+                    try:
+                        index = int(key)
+                        LOG.debug(f"key {key} is used as list index in {path}")
+                        path = path[index]
+                    except ValueError:
+                        path = path.setdefault(key, {})
```
mpvginde, line 97: Not sure if `try\except` close is the best thing to use here?

gmertes: The try\except clause is fine, but I would add another check that `path` is actually a list , because dictionary keys could also be integers. 

There is an edge case when integers are used as dict keys, and overriding a key in a non-existing entry. For example: `input.dataset.1.2=something`  will crash if `input.dataset.1` doesn't already exist as a dict in the config. In this case you want to create the `input.dataset={1: {2: ... }}` first with the setdefault.

Admittedly this is a very rare case, but I would still guard against it.

So I would split the logic in two:
1. Have the try\except to check if the key is an integer
2. check `type(path) is list`

Only if both 1 and 2 are true, use the key as a list index, otherwise call the `setdefault`

gmertes (changes requested):
Thanks for the contribution, this seems helpful and a nice solution to the linked issue.

Would you mind also adding a few lines in the documentation to describe this behaviour, maybe with an example? The docs are here:

https://github.com/ecmwf/anemoi-inference/blob/7923860abd237b463d035a5522975547b54f619d/docs/inference/apis/level3.rst?plain=1#L24-L27

src/anemoi/inference/config/__init__.py
```diff
@@ -89,7 +89,19 @@ def load(
...
+                            LOG.debug(f"key {key} is used as list index in list{path}")
+                            path = path[index]
+                        elif index == len(path):
+                            LOG.debug(f"key {key} is used to append to list {path}")
+                            path.append({})
+                            path = path[index]
+                        else:
+                            raise IndexError(f"Index {index} out of range for list {path} of length {len(path)}")
+                    else:
+                        path = path.setdefault(key, {})
```
mpvginde, line 104: @gmertes, I replaced the try/except clause to account for the edge cases and also support appending to the the list.

gmertes: @mpvginde Excellent! Could you also add some lines to the docs to document this behaviour? And then we are good to go.

gmertes (approved):

PR #211, opened 2025-04-25
fix: Allow for downstream tests to provide checkpoint mocks

HCookie:
Mocking checkpoints didn't allow for downstream tests to mock metadata


gmertes (approved):
LGTM

PR #216, opened 2025-04-28
feat: inference with external graph.

### LINKED ISSUE #215: Run inference with graph other than the one used in training.

A powerful feature of graph neural networks is that they can operate on any given graph. It is very interesting from an experimental and scientific perspective to be able to run inference using a graph that is different from the one used during training, as it allows us to investigate how well the models generalize to input that is not exactly the one they were exposed to during training. Example use cases are: use models trained as stretched grid to do forecasting as a lam model, or before transfer learning use the original model in inference on another domain/resolution to get an idea of the quality of the model without fine tuning. 

MetNo has such functionality implemented in their Bris inference:  https://github.com/metno/bris-inference/blob/038e8bdce3749210a4a4627e3b158edc61228214/bris/checkpoint.py#L185

The idea is to port this functionality to anemoi-inference by adding a new `external_graph` runner.

_No response_

### Additional context

_No response_

### Organisation

RMI

dietervdb-meteo:
## Description
We add functionality so that in inference the model can use another graph than the one it was trained with. In this first implementation the graph has to be provided as a file on disk. 

This PR adds a new runner `runner: external_graph` that is an extension of the default runner. The code is based on a similar feature in bris-inference: https://github.com/metno/bris-inference/blob/main/bris/checkpoint.py#L185

The runner can be selected and set in the config as follows:

```yaml
runner:
  external_graph:
    graph: path/to/graph.pt
```
For further options for the runner please consult the documentation.

## Type of Change

-   [x] Documentation update

## Issue Number

Closes #215 .

## Code Compatibility

### Dependencies

### Documentation

-   [x] I have updated the documentation and docstrings to reflect the changes
-   [x] I have added comments to my code, particularly in hard-to-understand areas

## Additional Notes

----
📚 Documentation preview 📚: https://anemoi-inference--216.org.readthedocs.build/en/216/


dietervdb-meteo:
Tested that using the 'external_graph' runner in the trivial setting, i.e. setting the graph that was used during training as external graph and letting the runner rebuild the model from scratch [including 'output_mask'], reproduces running the model using the default runner. E.g. 
mock_test
(RMSE against reanalysis, computed over all gridpoints and 36 dates)

src/anemoi/inference/config/run.py
gmertes, line 1: I suggest we don't add these entries to the main config, but pass them as options to the runner like this:

```yaml
runner:
  external_graph:
    graph: xxx
    output_mask: yyy
```

They then become keyword arguments in the constructor:
```python
class ExternalGraphRunner(DefaultRunner):
    def __init__(self, config, graph: str, output_mask: dict):
        ...
```

The downside is we don't get pydantic validation for these entries, but I would avoid bloating the config with optional entries for now. We can look at adding pydantic validation to the runner configs later. 

You can add the description of the options as a docstring in the constructor, then that will be included in the API docs.

src/anemoi/inference/runners/external_graph.py
```diff
@@ -0,0 +1,111 @@
...
+
+import torch
+
+from ..runners.default import DefaultRunner
+from . import runner_registry
+
+LOG = logging.getLogger(__name__)
+
+
+# Possibly move the function below to anemoi-models or anemoi-utils since it could be used in transfer learning.
```
gmertes, line 14: Good idea, but I'm happy to have a duplicate of this function here to get it working.

src/anemoi/inference/runners/external_graph.py
```diff
@@ -0,0 +1,111 @@
...
+        assert os.path.isfile(
+            graph_path
+        ), f"No graph found at {graph_path}. An external graph needs to be specified in the config file for this runner."
+        LOG.info(f"Loading external graph from path {graph_path}.")
+        return torch.load(graph_path, map_location="cpu", weights_only=False)
+
+    @cached_property
+    def model(self):
+        # load the model from the checkpoint
+        model_instance = super().model.to("cpu")
```
gmertes, line 96: This will load the model on GPU first before moving it to the CPU. 

You should be able to avoid the unnecessary GPU step, and load the model directly on the CPU by doing:
```python
device = self.device

self.device = "cpu"
model_instance = super().model
# rebuild model and inject weights 

self.device = device
return model_instance.to(self.device)

**Review by gmertes — CHANGES_REQUESTED** [review]
Thanks for this! 

The application is quite bespoke, and rebuilding the model like this is not guaranteed to be supported in the future. In fact, because it's doing some hacks like overwriting the supporting arrays this runner is likely to break in the future, especially without tests (but the upcoming integration tests could include this).

But since it is nicely compartmentalised in a runner I'm happy to accept it. 

Could you add a bit of documentation? You could add a new page under `usage/` titled "Running with external graph" or similar, with an example config and a bit of background, then put it in the index here:

https://github.com/ecmwf/anemoi-inference/blob/ba7a57606b91e8be2d0c2509c3d08f5a95d5c461/docs/index.rst#L118-L123

Title it "Running

File: src/anemoi/inference/runners/external_graph.py [hunk 3]
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, config):
+        super().__init__(config)
+        # Check if the external graph has the 'indices_connected_nodes' attribute
+        data = self.checkpoint._metadata._config.graph.data
+        assert data in self.graph.node_types, f"Node type {data} not found in external graph."
+        if "indices_connected_nodes" in self.graph[data]:
+            LOG.info(
+                "The external graph has the 'indices_connected_nodes' attribute."
+                " Patching metadata with MaskedGrid grid_indices."
+            )
```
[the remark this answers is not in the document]
dietervdb-meteo: When building a lam graph the boundary zone is essentially defined through a choice of edge builder in graph building. Almost by definition the boundary zone will not cover the entire globe and so there will be a (large) set of gridpoints (those outside the boundary zone) not connected to the processor mesh. They thus play no role in the model and it is thus best to remove them from the input. This is done through a postprocessor in the graph building, that drops all such unconnected nodes (See e.g. the default [lam graph config](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/config/graph/limited_area.yaml) for details). The subset of indices of those points in the dataset that in the end are part of the input is stored in a graph attribute called `indices_connected_nodes`. In training one should specify the use of this reduced input by passing `indices_connected_nodes` to the dataloader via its `grid_indices` attribute, see [here](https://github.com/ecmwf/anemoi-core/blob/1592d09f0915dddba7dcfc1c11897e0cae5cb6d0/training/src/anemoi/training/config/lam.yaml#L29). When the checkpoint is created during training the `indices_connected_nodes` are stored as a supporting array `grid_indices`, so that it  inference it can be used to select the correct input points (see [here](https://github.com/ecmwf/anemoi-inference/blob/fa9d5a87c896228aabd16b48baaacd8bad6fff08/src/anemoi/inference/inputs/dataset.py#L59)).

If we want to do inference on a lam graph with a model that was trained on a global graph then we need to add the `indices_connected_nodes` as `grid_indices` 'by hand', i.e. that is what this piece of code is about.

Do you suggest I add extra info to the log message, or a comment in the code? Or was it about a comment here on github?

src/anemoi/inference/runners/external_graph.py
dietervdb-meteo: The idea was to have the piece of code here showing it works and then later possibly port it (or a slight modification of it) to models/utils. See https://github.com/ecmwf/anemoi-core/issues/309

Shall I leave the comment in, or should it be removed?

src/anemoi/inference/config/run.py
dietervdb-meteo: That's a nice way to do it. Didn't work out of the box, but adapted runner entry of config [now allowed to be `dict` as well] and runner creation, so that now it does.

src/anemoi/inference/runners/external_graph.py
dietervdb-meteo: Good spot. Done.

src/anemoi/inference/runners/external_graph.py
```diff
@@ -0,0 +1,111 @@
...
+
+# Possibly move the function below to anemoi-models or anemoi-utils since it could be used in transfer learning.
+def inject_weights_and_biases(model, state_dict, ignore_mismatched_layers=False, ignore_additional_layers=False):
+    LOG.info("Updating model weights and biases by injection from an external state dictionary.")
+    # select weights and biases from state_dict
+    weight_bias_dict = {k: v for k, v in state_dict.items() if "bias" in k or "weight" in k}
+    model_state_dict = model.state_dict()
+
+    # check layers and their shapes
+    for key in weight_bias_dict.copy():
```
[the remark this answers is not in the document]
dietervdb-meteo: Actually you don't even need `.keys()`, `for key in list(weight_bias_dict):` works as well, implemented it that way. [Not putting `list` gives error as you probably know, as we are changing the dict while running through it.]

src/anemoi/inference/runners/external_graph.py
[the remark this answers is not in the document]
dietervdb-meteo: I added few words in comments + log message that might make things a little clearer.

dietervdb-meteo:
I added documentation where you suggested @gmertes . Might be a bit too prominent for this rather niche functionality? But don't see any other obvious place either. Feel free to suggest somewhere else, should be easy to move. Or we can leave it there for the moment.

src/anemoi/inference/runners/external_graph.py
```diff
@@ -0,0 +1,127 @@
...
+from . import runner_registry
+
+LOG = logging.getLogger(__name__)
+
+
+# Possibly move the function below to anemoi-models or anemoi-utils since it could be used in transfer learning.
+def inject_weights_and_biases(model, state_dict, ignore_mismatched_layers=False, ignore_additional_layers=False):
+    LOG.info("Updating model weights and biases by injection from an external state dictionary.")
+    # select weights and biases from state_dict
+    weight_bias_dict = {k: v for k, v in state_dict.items() if "bias" in k or "weight" in k}
```
havardhhaugen, line 18: I don't think the pre_processors and post_processors layers are transferred correctly using this method, since they do not contain bias or weight in their name. When I replace the graph and call '_build_model()' they change (not sure if they are initialized randomly), but they also need to be copied over in this function.

dietervdb-meteo: Thanks for pointing that out. For the checkpoints I considered so far there was no difference in these layers. I will have a look to see if I can reproduce your issue and then see how to fix it, as we discussed.

src/anemoi/inference/runners/external_graph.py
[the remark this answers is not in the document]
gmertes: Yes just as a comment in the code, thanks!

src/anemoi/inference/config/run.py
gmertes: Oh my bad. That functionality and your adaptation already exists in a function but it wasn't used yet for the runner. I changed it and pushed it to the branch, can you check that it still works for you?

src/anemoi/inference/runners/external_graph.py
gmertes: Keeping the comment is fine 👍

gmertes (commented):
Thanks for the docs! Yes it's a bit prominent for a niche feature but it's the most logical place we have for now, let's keep it there and improve it later.

src/anemoi/inference/config/run.py
dietervdb-meteo: Great, that works.

src/anemoi/inference/runners/external_graph.py
dietervdb-meteo: I double checked and the normalization layers are actually fine in the current implementation, even when running from a different dataset. Still to protect from crazy stuff happening in use cases I might have overlooked I implemented a check to see if the normalization layers are fine in the end. See: https://github.com/ecmwf/anemoi-inference/blob/2259ccffa5c4c08150c1f749c29464b22ad4aaf8/src/anemoi/inference/runners/external_graph.py#L176

frazane:
Wait before merging - I am working on adding support of this feature also to inference from GRIB (possibly others) files inputs. I got it to work but I still need to figure some things (e.g. I had to remove some shape checks in the runner).

I'll push my commits early next week!

src/anemoi/inference/runners/external_graph.py
havardhhaugen: Turns out this issue was exclusive to transferlearned / finetuned models, where model.statistics does not always match the  pre and post processors in the finetuned checkpoint (causing the issue with build_model() updating to the wrong post and pre-processors). This should probably be addressed in anemoi-training too, but for now we should implement functionality to transfer these layers since we already have finetuned checkpoints with this issue.

gmertes (commented):
Thanks @frazane for the changes!

@dietervdb-meteo you are still working on some extra functionality related to the processors? Let me know when the PR is ready, to me it looks good. Were there any comments from ATS or was it all good?

gmertes:
No worries, I believe the merge will only be blocked if I was the last committer. Of course I'm happy for Havard to also give it a review.

PR #220, opened 2025-04-30
test: integration test set up

VeraChristina:
## Description
 
A few small changes for the basic set up for integration tests:
- Move unit tests to subfolder tests/unit
- Add subfolder tests/integration
- Add conftest in tests directory, to add the --longtest option for slow integration tests

I made a small change to mock_load_metadata in inference/testing/mock_checkpoint.py so that it does not assume the previous folder structure. The configs now need to specify paths relative to the tests/ directory.
Alternative: leave the config and and checkpoints folders at the top level of tests/ -- that would make more sense if we intend to reuse any of these for integration tests.


theissenhelen (approved):
Thanks for this!

gmertes (approved):
I think having the configs and checkpoints separate for unit and integration test is a good idea for now, so that the work on integration tests doesn't impact the unit and vice versa.

If we see a lot of overlap down the line, we can merge them back into a shared space.

Calling @HCookie for a second opinion.

HCookie (dismissed):

HCookie (approved):
LGTM

PR #222, opened 2025-05-05
fix!: do not set 'eps' key in grib encoding

frazane:
## Description

During the encoding of the `grib` output, `eccodes` raises the following error:

```
KeyValueNotFoundError: Key/value not found
```

this happens when trying to set the `eps` key. It appears that this key not universally compatible with GRIB encodings. Therefore, this PR removes the hardcoded specification. 

## Type of Change

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

### DIFF
--- a/src/anemoi/inference/grib/encoding.py
+++ b/src/anemoi/inference/grib/encoding.py
@@ -246,8 +246,6 @@ def grib_keys(
 
     result["edition"] = edition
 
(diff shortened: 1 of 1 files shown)


gmertes (approved):
Thanks!

PR #227, opened 2025-05-06
feat: temporal interpolation runner

radiradev:
## Description
This is a draft PR to receive some feedback on the implementation of the interpolator inference before I proceed with further changes. 

Currently `self.create_target_forcings`  uses both the `input_state` and the `input_tensor_torch` - this can be refactored by inferring the shape of the forcings directly. 

I also think the `self.forecast` method should be implemented differently. It currently expects a `lead_time` argument but this is not used and perhaps _forecast_ is not the best name for this task and should be something more general. Perhaps we should change the `run()` method of the base runner?

It is yet to be implemented for parallel inference and using a forecasting model to produce the target forcings during inference.

## Type of Change

-   [ ] Documentation update

## Issue Number

Since https://github.com/ecmwf/anemoi-core/commit/ea644ce1c9aef902333d9cbb30bcde0a3746fbcc introduced functionality for interpolating between two weather states (e.g  given 0h, +6h model outputs (+1h, +2h ..., +5h)). Changes are required in `anemoi-inference` as well to accommodate for this new functionality. 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/anemoi/inference/inputs/dataset.py
+++ b/src/anemoi/inference/inputs/dataset.py
@@ -117,7 +117,19 @@ class DatasetInput(Input):
         fields = input_state["fields"]
 
         date = np.datetime64(date)
-        data = self._load_dates([date + np.timedelta64(h) for h in self.checkpoint.lagged])
+
+        # We should be able to use `explicit_times` in this way for every model in the future
+        # However it was introduced with the interpolator 
+        # Thus we keep the else branch for backward compatibility
+        if hasattr(self.checkpoint._metadata._config_training, 'explicit_times'):
+            timestep = self.checkpoint._metadata.timestep
+            input_explicit_times = self.checkpoint._metadata._config_training.explicit_times.input
+            dates = [date + (h * timestep) for h in input_explicit_times]
+        
+        else:
+            dates = [date + np.timedelta64(h) for h in self.checkpoint.lagged]
+
+        data = self._load_dates(dates)
 
         if data.shape[2] != 1:
             raise ValueError(f"Ensemble data not supported, got {data.shape[2]} members")
--- a/src/anemoi/inference/runners/interpolator.py
+++ b/src/anemoi/inference/runners/interpolator.py
@@ -1,0 +1,223 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import torch
+import logging
+from typing import Any
+from typing import Dict
+from typing import List
+import numpy as np
+
+from anemoi.inference.runner import Kind
+
+from anemoi.utils.timer import Timer
+from anemoi.utils.dates import frequency_to_timedelta as to_timedelta
+from anemoi.inference.types import State
+import torch.nn.functional
+
+from ..forcings import ComputedForcings
+from ..forcings import Forcings
+from .simple import SimpleRunner
+from . import runner_registry
+from ..profiler import ProfilingLabel
+from typing import Tuple
+import datetime
+
+
+from typing import Generator
+
+LOG = logging.getLogger(__name__)
+
+
+@runner_registry.register("interpolator")
+class InterpolatorRunner(SimpleRunner):
+    """Use that runner when using the low level API."""
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the InterpolatorRunner.
+
+        Parameters
+        ----------
+        *args : tuple
+            Positional arguments.
+        **kwargs : dict
+            Keyword arguments.
+        """
+        super().__init__(*args, **kwargs)
+        self.target_forcings = self.target_computed_forcings(self.checkpoint._metadata._config_training.target_forcing.data)
+        # check that model is an Interpolator
+    
+    def predict_step(self, model: torch.nn.Module, input_tensor_torch: torch.Tensor, target_forcing: torch.Tensor) -> torch.Tensor:
+        return model.predict_step(input_tensor_torch, target_forcing)
+
+    
+    def target_computed_forcings(self, variables: List[str], mask=None) -> List[Forcings]:
+        """Create dynamic computed forcings.
+
+        Parameters
+        ----------
+        variables : List[str]
+            The variables for the forcings
+
+        Returns
+        -------
+        List[Forcings]
+            The created dynamic computed forcings.
+        """
+        result = ComputedForcings(self, variables, mask)
+        LOG.info("Dynamic computed forcing: %s", result)
+        return [result]
+
+    def interpolator_stepper(
+        self, start_date: datetime.datetime
+    ) -> Generator[Tuple[datetime.timedelta, datetime.datetime, datetime.datetime, bool], None, None]:
+        """Generate step and date variables for the forecast loop.
+
+        Parameters
+        ----------
+        start_date : datetime.datetime
+            Input start date
+            
+        Returns
+        -------
+        step : datetime.timedelta
+            Time delta since beginning of forecast
+        valid_date : datetime.datetime
+            Date of the forecast
+        next_date : datetime.datetime
+            Date used to prepare the next input tensor
+        is_last_step : bool
+            True if it's the last step of the forecast
+        """
+        target_steps = self.checkpoint._metadata._config_training.explicit_times.target
+        steps = len(target_steps)
+
+        LOG.info("%s, time stepping: %s Interpolating %s steps", self.checkpoint.timestep, steps)
+
+        for s in range(steps):
+            step = target_steps[s]  * self.checkpoint.timestep
+            date = start_date + step
+            is_last_step = s == steps - 1
+            yield step, date, target_steps[s], is_last_step
+
+    def create_target_forcings(
+        self, dates, state: State, input_tensor_torch: torch.Tensor, interpolation_step: int
+    ) -> torch.tensor:
+        """Create target forcings tensor.
+
+        Parameters
+        ----------
+        date : datetime.datetime
+            The date.
+
+        state : State
+            The state dictionary. 
+
+        Returns
+        -------
+        torch.Tensor
+            the target forcings.
+            
+        """
+        batch_size, _, grid, n_vars = input_tensor_torch.shape
+
+        # Should use self.checkpoint._metadata._config_training.target_forcing.time_fraction but not accessible
+        use_time_fraction = True
+
+        # 1 for ensemble size but not needed for the other input??
+        target_forcings = torch.empty(batch_size, 1, grid, len(self.target_forcings) + use_time_fraction, device=input_tensor_torch.device, dtype=input_tensor_torch.dtype)
+        for idx, source in enumerate(self.target_forcings):
+            arrays = source.load_forcings_array(dates, state)
+            for name, forcing in zip(source.variables, arrays):
+                assert isinstance(forcing, np.ndarray), (name, forcing)
+                target_forcings[..., idx]  = torch.tensor(forcing, device=input_tensor_torch.device, dtype=input_tensor_torch.dtype)
+                self._input_kinds[name] = Kind(forcing=True, constant=True, **source.kinds)
+                if self.trace:
+                    self.trace.from_source(name, source, "target forcings")
+
+        if use_time_fraction:
+            boundary_times = self.checkpoint._metadata._config_training.explicit_times.input
+            # this only works with two boundary times?
+            target_forcings[..., -1] = (interpolation_step - boundary_times[-2]) / (
+                    boundary_times[-1] - boundary_times[-2]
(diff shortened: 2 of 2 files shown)


src/anemoi/inference/inputs/dataset.py
```diff
@@ -117,7 +117,19 @@ def create_input_state(self, *, date: Optional[Date] = None) -> State:
...
 
         date = np.datetime64(date)
-        data = self._load_dates([date + np.timedelta64(h) for h in self.checkpoint.lagged])
+
+        # We should be able to use `explicit_times` in this way for every model in the future
+        # However it was introduced with the interpolator
+        # Thus we keep the else branch for backward compatibility
+        if hasattr(self.checkpoint._metadata._config_training, "explicit_times"):
+            timestep = self.checkpoint._metadata.timestep
+            input_explicit_times = self.checkpoint._metadata._config_training.explicit_times.input
```
HCookie, line 126: Instead of accessing the private attributes of the checkpoint, I'd suggest you add these variables to the `Checkpoint` and `Metadata` class.

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,232 @@
...
+from ..profiler import ProfilingLabel
+from . import runner_registry
+from .simple import SimpleRunner
+
+LOG = logging.getLogger(__name__)
+
+
+@runner_registry.register("interpolator")
+class InterpolatorRunner(SimpleRunner):
+    """Use that runner when using the low level API."""
```
HCookie, line 37: Please update the docs

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,232 @@
...
+        *args : tuple
+            Positional arguments.
+        **kwargs : dict
+            Keyword arguments.
+        """
+        super().__init__(*args, **kwargs)
+        self.target_forcings = self.target_computed_forcings(
+            self.checkpoint._metadata._config_training.target_forcing.data
+        )
+        # check that model is an Interpolator
```
HCookie, line 53: Is this a TODO?

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,232 @@
...
+        ----------
+        *args : tuple
+            Positional arguments.
+        **kwargs : dict
+            Keyword arguments.
+        """
+        super().__init__(*args, **kwargs)
+        self.target_forcings = self.target_computed_forcings(
+            self.checkpoint._metadata._config_training.target_forcing.data
+        )
```
HCookie, line 52: How expensive is this compute? Would it make sense to delay after the init.

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,232 @@
...
+
+        Parameters
+        ----------
+        date : datetime.datetime
+            The date.
+
+        state : State
+            The state dictionary.
+
+        Returns
```
HCookie, line 122: Can you add all the params?

HCookie (changes requested):

anaprietonem:
@Rilwan-Adewoyin @mc4117 - if you have a temporal interpolator model trained would it be possible for you to test this PR?

Rilwan-Adewoyin:
> @Rilwan-Adewoyin @mc4117 - if you have a temporal interpolator model trained would it be possible for you to test this PR?

Will provide an update when this has been tested

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,254 @@
...
+        Parameters
+        ----------
+        *args : tuple
+            Positional arguments.
+        **kwargs : dict
+            Keyword arguments.
+        """
+        super().__init__(*args, **kwargs)
+
+        assert isinstance(
```
Rilwan-Adewoyin, line 55: When running from CLI with a config.yaml - this line `self.model` causes the error below:

 This happens because you subclass SimpleRunner instead of DefaultRunner - where SimpleRunner is compatible with your pythonAPI based approach whereas DefaultRunner works when you have a config.yaml that defines the args/kwargs 

(@HCookie @gmertes  is my understanding correct? and if-so are there plans to merge the two initialisation strategies together into one class that supports both. The pythonAPI approach has advantages when it comes to interactive debugging ) 



```
2025-05-15 07:58:35 INFO Using InterpolatorRunner runner, device=cuda
Traceback (most recent call last):
  File "[redacted]bin/anemoi-inference", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "[redacted]/temporal_interpolator/src/anemoi/inference/__main__.py", line 38, in main
    cli_main(__version__, __doc__, COMMANDS)
  File "[redacted]/pyenv/versions/aifs_py311/lib/python3.11/site-packages/anemoi/utils/cli.py", line 229, in cli_main
    cmd.run(args)
  File "[redacted]/temporal_interpolator/src/anemoi/inference/commands/run.py", line 59, in run
    runner = create_runner(config)
             ^^^^^^^^^^^^^^^^^^^^^
  File "[redacted]/temporal_interpolator/src/anemoi/inference/runners/__init__.py", line 33, in create_runner
    return runner_registry.create(config.runner, config, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[redacted]/pyenv/versions/aifs_py311/lib/python3.11/site-packages/anemoi/utils/registry.py", line 315, in create
    return factory(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "[redacted]/temporal_interpolator/src/anemoi/inference/runners/interpolator.py", line 56, in __init__
    self.model, AnemoiModelEncProcDecInterpolator
    ^^^^^^^^^^
  File "[redacted]/.pyenv/versions/3.11.8/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "[redacted]/temporal_interpolator/src/anemoi/inference/runner.py", line 450, in model
    with Timer(f"Loading {self.checkpoint}"):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[redacted]/temporal_interpolator/src/anemoi/inference/checkpoint.py", line 104, in __repr__
    return f"Checkpoint({self.path})"
                         ^^^^^^^^^
  File "[redacted]/.pyenv/versions/3.11.8/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "[redacted]/temporal_interpolator/src/anemoi/inference/checkpoint.py", line 122, in path
    raise TypeError(f"Cannot parse model path: {path}. It must be a path or dict")
TypeError: Cannot parse model path: description=None checkpoint='/ec/res4/scratch/ecm9156/aifs/checkpoint/e31139ce71b64760b9964a1cc4ff367a/inference-last.ckpt' runner='interpolator' date=datetime.datetime(2023, 6, 1, 0, 0) lead_time=240 name=None verbosity=1 use_profiler=False world_size=1 report_error=False input='mars' output={'grib': '/ec/res4/scratch/ecm9156/aifs/inference_output/e31139ce71b64760b9964a1cc4ff367a/output.grib'} pre_processors=[] post_processors=None forcings=None device='cuda' precision=None allow_nans=None use_grib_paramid=False write_initial_state=True output_frequency=None env={} patch_metadata={} development_hacks={} trace_path=None debugging_info={}. It must be a path or dict
```

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,254 @@
...
+            self.model, AnemoiModelEncProcDecInterpolator
```
Rilwan-Adewoyin, line 56: This assertion is currently incorrect, it should read:

`self.model.model, AnemoiModelEncProcDecInterpolator`

confusing but, self.model is always an anemoi.models.interface class and self.model.model will 
 be the anemoi.models.models class

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,254 @@
...
+            Date of the zeroth index of the input tensor
+        target_index : int
+            Date used to prepare the next input tensor
+        is_last_step : bool
+            True if it's the last step of interpolation
+        """
+        target_steps = self.checkpoint._metadata._config_training.explicit_times.target
+        steps = len(target_steps)
+
+        LOG.info("%s, time stepping: %s Interpolating %s steps", self.checkpoint.timestep, steps)
```
Rilwan-Adewoyin, line 109: Not enough args for the # of %s : I think the %s, is not meant to be there?

src/anemoi/inference/runners/interpolator.py
```diff
@@ -0,0 +1,254 @@
...
+        if self.verbosity > 0:
+            self._print_input_tensor("First input tensor", input_tensor_torch)
+
+        for s, (step, date, interpolation_step, is_last_step) in enumerate(self.interpolator_stepper(start)):
+            title = f"Interpolating step {step}({date})"
+
+            # this should be changed
+            result["date"] = date
+            result["previous_step"] = result.get("step")
+            result["step"] = self.checkpoint.timestep
```
Rilwan-Adewoyin, line 221: should this not be `result["step"] = step`

gmertes: Agreed that seems incorrect

src/anemoi/inference/runners/interpolator.py
gmertes: We do need to think about that class structure, but let's put that out of scope of this PR. 
The interpolator should inherit from `DefaultRunner`. 

If you then want to instantiate the interpolator from code without a config file, you use the [RunConfiguration.load](https://github.com/ecmwf/anemoi-inference/blob/748226d454ecf3b6848b40e75fbc55cf2421a30f/src/anemoi/inference/commands/retrieve.py#L169). This function can take a plain dictionary.

gmertes (commented):
This already looks really nice.

On the comment of `self.forecast` , we could rename that to something like `self.predict` but ultimately it's just semantics.

radiradev:
@mchantry Thanks! As this PR does not tackle interpolating in parallel with a forecaster I think updating the _true_ (or one given by first doing inference with a forecaster) state should be done outside of the runner. So to iterate over dates one could currently do:

```
runner = SimpleRunner(checkpoint)
args, kwargs = runner.checkpoint.open_dataset_args_kwargs(
    use_original_paths=True, from_dataloader="test"
)
input = DatasetInput(runner, args, kwargs)
for date in dates:
    input_state = input.create_input_state(date=date)
    runner.run(...)
   ```
  I agree this is not ideal, but it is inline with the `anemoi-inference` design choice that the _true_ state is not updated within the runner.

radiradev:
Hi @gmertes thanks for the comment. I see, I will update the runner to inherit from `DefaultRunner`. 

I agree that it would be great to have the outer loop through lead times so it can be used on a file previously inferred with a forecaster. However, I am unsure how this should be implemented, as for the forecaster - only one state is passed to the model and then the model uses its own predictions to update that state: https://github.com/ecmwf/anemoi-inference/blob/bab2dbe70f1d7a7739b0228f93dd4632e8f8d41d/src/anemoi/inference/runner.py#L588-L613

Do let me know if you have any pointers on how this should be implemented.

Rilwan-Adewoyin:
> Thanks for the fixes @Rilwan-Adewoyin! On your discussion points, I would support having diagnostic capability in the timeinterpolator. How difficult would this be to fix?

Yep, so to clarify - anemoi-inference can still be done with a time interpolator model that has been trained with diagnostics, and an output saved with no issues. I think a slight point of concern is any code that ingests the grib file must be aware that the output dataset will be jagged. I'm not aware if any downstream workflows rely on a regular shaped array being able to be built from the individual messages in the grib

Rilwan-Adewoyin:
Most recent set of commits addresses the following 3 concerns raised by @gmertes 

1.  2301528d13e7866840c3a4fc61233215a24d078c - reimplements the changes to the checkpoint.lagged property required for TimeInterpolator to run as a patching that occurs during init of the TimeInterpolator class
2. c910d5edc831b43c484688a6a88a829908b1449e - moves the custom init logic that existed in the TimeInterpolator. 
3. ab94ed7386693faf59cf10ed72ced8860164ebf3 - simplifies the state saving logic involved when producing windows of predictions that contribute to a final contiguous sequence

Rilwan-Adewoyin:
On the most recent set of commits:

- The bulk are because I had to edit an old commit from @radiradev to fix the author details to pass cla check

Some of the most recent set of commits addresses the following 3 concerns raised by @gmertes 

1.  eaf10923d57571c892f69e7c22f7d0a6a3392619 - reimplements the changes to the checkpoint.lagged property (required for TimeInterpolator to run) as a patching that occurs during init of the TimeInterpolator class
2. 7822a0884f16f2871f8b586c2a3dc0723e395e72 - moves the custom init logic that existed in the TimeInterpolator so TimeInterpolator has a more standard init logic 
3. 43a2981354ce43f342212a09ef6aec743cb88a5c - simplifies the state saving logic involved when producing windows of predictions that contribute to a final contiguous sequence

Rilwan-Adewoyin (approved):

gmertes (dismissed):

gmertes (approved):

PR #231, opened 2025-05-16
fix: rename method on ForwardOutput class

frazane:
## Description

Rename the method `write_initial_step` of the `ForwardOutput` class to the correct name `write_initial_state`.


gmertes (approved):
Thanks!

PR #233, opened 2025-05-21
feat(grib output): support string format specification for output filename

frazane:
## Description

Add the ability to use string format specifiers (such as zero-padding) in the GRIB output path template, for instance:
```yaml
output:
  grib:
    path: output/grib/{dateTime}_{step:03}.grib
```

will give outputs like `output/grib/{dateTime}_006.grib`, `output/grib/{dateTime}_012.grib`, etc.

----
📚 Documentation preview 📚: https://anemoi-inference--233.org.readthedocs.build/en/233/


gmertes (approved):
Great work, thanks for adding the docs and test!

PR #236, opened 2025-05-28
feat: Mars with Nested models

HCookie:
## Description
The mars input and by association other inputs cannot be used with nested models.
See #235 for context.

In addition to the changes in this PR, some extra manual patches of the checkpoint are required. 

This enables an input of 
```yaml
cutout:
	lam_0:
		mars:
			grid: '0.01/0.01'
			area: '2/-2/-2/2'
		mask: 'source0/trimedge_mask'
	global: 'mars'
```

Using #247, can be tested with external graphs

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

PR #249, opened 2025-06-13
fix: apply mask to all sources in cutout input.

### LINKED ISSUE #248: First mask of cutout input is not applied
### What happened?

In the Cutout input file, the mask of the first source is never applied.

For example with the following config:
```yaml
input:
  cutout:
    global:
        grib: path/to/global.grib
    lam_0:
        grib: path/to/lam_0.grib
```

The cutout mask of `global` is never applied. 
Note that in this instance, reversing the order of global and lam_0 would fix it, but there could be settings with multiple lam.
This also differs from the implementation in anemoi-datasets where the mask is applied everywhere (e.g. [here](https://github.com/ecmwf/anemoi-datasets/blob/f9769d7944738ecbedb6b3cc1f78cd26de36a73f/src/anemoi/datasets/data/grids.py#L544)).

I will provide an initial PR with a fix applying the mask everywhere.

Make an inference 

### Version

main

Linux ac6-102.bullx 4.18.0-477.43.1.el8_8.x86_64 # 1 SMP Thu Jan 4 10:30:57 EST 2024 x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

Météo-France

bdvllrs:
## Description
In the Cutout input file, the mask of the first source is never applied.

Previously, the mask of the first source in config.input.cutout was not applied.
For example with the following config:
```yaml
input:
  cutout:
    global:
        grib: path/to/global.grib
    lam_0:
        grib: path/to/lam_0.grib
```

The cutout mask of `global` is never applied. 
This applies the masks to all sources.

Fixes #248

##  Additional notes ##

### DIFF
--- a/src/anemoi/inference/inputs/cutout.py
+++ b/src/anemoi/inference/inputs/cutout.py
@@ -9,15 +9,13 @@
 
 
 import logging
-from typing import List
-from typing import Optional
+from typing import Any, List, Optional
 
 import numpy as np
 
 from anemoi.inference.input import Input
 from anemoi.inference.inputs import create_input
-from anemoi.inference.types import Date
-from anemoi.inference.types import State
+from anemoi.inference.types import Date, State
 
 from . import input_registry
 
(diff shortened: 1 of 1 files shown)


HCookie:
Hi, 
Good spot, that is definitely something we should address.
I had made some changes in #236 to this functionality, and may help to provide a cleaner solution

bdvllrs:
Ok, how do you want to work on this. Should I wait for #236 to be merged first, or should we add this directly to your PR?

bdvllrs:
@HCookie Yep, I've just seen, I will rebase my PR. This should be simpler than what I did with your refactor.

HCookie (dismissed):

HCookie:
@dietervdb-meteo @frazane Do you have a chance to look this over?

bdvllrs:
Following @frazane comment, I started from scratch to keep the same method to respect the performance requirement. Tell me if that's ok with you.

Note: Even if the mask of the first lam is not strictly necessary, I still think applying all masks is the best solution because it avoids any bug due to input order in the cutout input.

frazane:
>Note: Even if the mask of the first lam is not strictly necessary, I still think applying all masks is the best solution because it avoids any bug due to input order in the cutout input.

I think this is a sensible choice. Especially considering that in the future the "ordering" might change or even not be a thing. All good for me!

bdvllrs:
So... I've added a test to actually check what I was saying and there is actually a slight bug of dimensions whith more than 2 sources (None is adding a dimension to the state). I'm opening a new PR with a test and the fix.

Issue #253, opened 2025-06-18
apply_mask output is broken


dietervdb-meteo:
### What happened?

```
in __init__
    mask=self.checkpoint.load_supporting_array(mask),
    mask=self.checkpoint.load_supporting_array(mask),
    mask=self.checkpoint.load_supporting_array(mask),
    mask=self.checkpoint.load_supporting_array(mask),
    mask=self.checkpoint.load_supporting_array(mask),
AttributeError: 'ApplyMaskOutput' object has no attribute 'checkpoint'
```

### What are the steps to reproduce the bug?

put 
``` yaml
output:
  apply_mask:
    mask: favorite/mask/in/checkpoint
```
in the checkpoint

### Version

latest main

### Platform (OS and architecture)

irrelevant

### Relevant log output

```shell

```

### Organisation

RMI

dietervdb-meteo:
Hi, see your comment just now. Already went ahead and fixed it in another way, by calling the mask from the checkpoint in the super masked, where the output is already instantiated and has the checkpoint attribute. This slightly changes the functionality of masked, but too me looks like it enhances it a little, since now one can get the mask automatically from the checkpoint in case it is not explicitly provided. Anyways, have a look at the PR, we can continue discussion there.

gmertes:
fixed in #254

PR #254, opened 2025-06-18
fix: apply_mask output

dietervdb-meteo:
## Description

#253 

fixes the apply_mask output

#253


HCookie:
Could you consolidate the `extract_lam` output as well?

gmertes (approved):
Many thanks Dieter!

PR #260, opened 2025-06-27
feat: inner-level processors

### LINKED ISSUE #251: Different pre-processors for for different input sources in cutout

Currently, the preprocessors are created by the runner and they are used indifferently on all input sources.

This is problematic in the case of a cutout where we might want different pre-processors for the different sources (e.g. use remove_nans for the local source and nothing for the global source).

I would happily provide a PR with the chosen method.

Because pre-processors are applied by inputs, maybe pre-processors should rather be a parameter of inputs?

For example for a cutout input with two grib inputs:
```yaml
input:
  cutout:
    lam_0:
        grib:
          path: /path/to/local.grib
          pre_processors:
            - forward_transform_filter: remove_nans
    global:
        grib:
          path: /path/to/global.grib
```

This would also make it explicit that preprocessing are handled by inputs.

An alternative would be to have preprocessors prefixed by a source:
```yaml
pre_processors:
    lam_0:
    - forward_transform_filter: remove_nans
```
It could be defined in the `create_pre_processors` of a cutout runner.

(I have seen that there is a current Cutout runner, but I haven't yet understood its purpose in the context of a cutout inference).

### Additional context

_No response_

### Organisation

Météo-France

bdvllrs:
## Description
Introduces input-level pre-processors and output-level post-processors.

Currently, the pre-processors are created by the runner and they are used indifferently on all input sources.

This is problematic in the case of a cutout where we might want different pre-processors for the different sources (e.g. use remove_nans for the local source and nothing for the global source).

Similarly for post-processors when using the `tee` output.

Fixes #251.

##  Additional notes ##

### DIFF
--- a/docs/inference/configs/top-level.rst
+++ b/docs/inference/configs/top-level.rst
@@ -83,6 +83,13 @@ and output. It set to ``null`` (default), the value is set internally to
 The entries for the inputs and outputs are specified in the :ref:`inputs
 <inputs>` and :ref:`outputs <outputs>` sections of the documentation.
 
+**************************
+ Pre- and Post-processors
+**************************
+
+The entries for the pre- and post-processors are specified in the
+:ref:`processors <processors>` section of the documentation.
+
 ***********
  Debugging
 ***********
--- a/src/anemoi/inference/runners/default.py
+++ b/src/anemoi/inference/runners/default.py
@@ -96,7 +96,6 @@ class DefaultRunner(Runner):
         input = self.create_input()
         output = self.create_output()
 
-        # pre_processors = self.pre_processors
         post_processors = self.post_processors
 
         input_state = input.create_input_state(date=self.config.date)
--- a/src/anemoi/inference/runners/default.py
+++ b/src/anemoi/inference/runners/default.py
@@ -96,24 +96,17 @@ class DefaultRunner(Runner):
         input = self.create_input()
         output = self.create_output()
 
-        post_processors = self.post_processors
-
         input_state = input.create_input_state(date=self.config.date)
 
         # This hook is needed for the coupled runner
         self.input_state_hook(input_state)
 
         state = Output.reduce(input_state)
-        for processor in post_processors:
-            state = processor.process(state)
 
         output.open(state)
         output.write_initial_state(state)
 
         for state in self.run(input_state=input_state, lead_time=lead_time):
-            for processor in post_processors:
-                LOG.info("Post processor: %s", processor)
-                state = processor.process(state)
             output.write_state(state)
 
         output.close()
--- a/docs/inference/configs/introduction.rst
+++ b/docs/inference/configs/introduction.rst
@@ -49,6 +49,7 @@ or if it expects a single argument, it can be simplified as:
    top-level
    inputs
    outputs
+   processors
    forcings
    icon-input
    grib-input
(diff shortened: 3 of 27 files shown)


HCookie (changes requested):
Thanks for the PR. It matches the overall OOD.
However, I think I disagree with applying the top level `pre_processors` in each input. Consider the `cutout` input, it makes more sense to apply the top level processors to the combined input state rather than on each component.

src/anemoi/inference/runners/default.py
```diff
@@ -96,25 +96,17 @@ def execute(self) -> None:
...
-        for processor in post_processors:
-            state = processor.process(state)
 
         output.open(state)
         output.write_initial_state(state)
 
         for state in self.run(input_state=input_state, lead_time=lead_time):
-            for processor in post_processors:
-                LOG.info("Post processor: %s", processor)
-                state = processor.process(state)
```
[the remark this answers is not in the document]
bdvllrs: I agree that in practice this makes more sense, however input state are applied on `ekd.FieldList` and not `State` so they can't be easily applied from outside the inputs. Indeed, state["fields"] is not a FieldList but a dict of numpy arrays.

Note that pre-processors were already applied in the ekd input (and only there) before (I assume because of this).

There is however no problem to do the post-processors in `execute` as they already expect a `State`.

bdvllrs:
As explained as a response to @gmertes comment (https://github.com/ecmwf/anemoi-inference/pull/260#discussion_r2189553807), this reasoning would be ideal. However because of the current structure, this makes it complicated for pre-processors to be applied in the runner. Maybe we can work on a refactoring of pre-processors in a future PR?

gmertes (commented):
Ah yes, sorry, I forgot that that is indeed the reason why pre processors are applied in the input. I agree to keep it like this, we could consider refactoring it in another PR but it's also more practical to do pre-processing on the FIeldList instead of the State/numpy.

src/anemoi/inference/runners/default.py
```diff
@@ -97,26 +97,21 @@ def execute(self) -> None:
...
-
         input_state = input.create_input_state(date=self.config.date)
 
         # This hook is needed for the coupled runner
         self.input_state_hook(input_state)
 
         state = Output.reduce(input_state)
-        for processor in post_processors:
-            LOG.info("Post processor: %s", processor)
-            state = processor.process(state)
```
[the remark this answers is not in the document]
bdvllrs: Oh yes, my bad! I've added it back in 9e8936d.

docs/inference/configs/top-level.rst
```diff
@@ -83,6 +83,13 @@ and output. It set to ``null`` (default), the value is set internally to
 The entries for the inputs and outputs are specified in the :ref:`inputs
 <inputs>` and :ref:`outputs <outputs>` sections of the documentation.
 
+**************************
+ Pre- and Post-processors
+**************************
+
+The entries for the pre- and post-processors are specified in the
+:ref:`processors <processors>` section of the documentation.
```
gmertes, line 91: Sorry found one last issue here and then we are good to merge! 
This link is somehow linking to the anemoi-training docs 🤔

bdvllrs: I forgot to add the label in processors and I added "inference-processors". It works on my local build of the docs, but it seems like the online build is not taking this into account, I'm a bit confused.

gmertes: Something to do with how all the anemois are in the same readthedocs project, interesting that it's automatically cross-linking.

gmertes (approved):
Awesome work, many thanks!

PR #282, opened 2025-07-18
fix(grib): Ocean grib encoding

gmertes:
## Description
A three-for-one grib special:

- Remove the hardcoded edition=1 for hindcasts so that we can do hindcasts with grib2-only variables, like those in the ocean. The edition comes from the template, so grib1 in will still do grib1 out. 
- Perform time encoding with the period information in the variable metadata, instead of the model's step size. This is to support fields that have different periods than the model step (like ocean averages).
- When the `accumulate_from_start_of_forecast` processor was enabled, we didn't properly set `startStep=0`, it was doing `startStep=previous_step`.

Draft for now (tests will be failing) because it needs a new release of transform after merging this:
https://github.com/ecmwf/anemoi-transform/pull/132

### DIFF
--- a/src/anemoi/inference/grib/encoding.py
+++ b/src/anemoi/inference/grib/encoding.py
@@ -10,6 +10,7 @@
 
 import logging
 import re
+import warnings
 from io import IOBase
 from typing import TYPE_CHECKING
 from typing import Any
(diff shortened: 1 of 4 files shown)


src/anemoi/inference/grib/encoding.py
```diff
@@ -152,7 +158,22 @@ def encode_time_processing(
...
         previous_step = step
 
-    start = _step_in_hours(start_steps.get(variable, previous_step))
+    if period := getattr(variable, "period", None):
+        start = step - period
+        if start < as_timedelta(0):
+            raise ValueError(
+                f"Writing {variable.name} at step {_step_in_hours(step)} with period {_step_in_hours(period)} for would result in a negative start step {_step_in_hours(start)}. "
+                "Try `write_initial_state: False`"
+            )
```
HCookie, line 167: I think precommit should be getting mad here, long error message in raise.
https://docs.astral.sh/ruff/rules/f-string-in-exception/

HCookie (approved):

PR #297, opened 2025-08-04
feat(runner): Dynamically evalute default device from config

### LINKED ISSUE #296: Generalise device assignment default
### What happened?

The default values in the `Runner` and `Config` assume `cuda` as the gpu environment. As we do not attempt to assign a number of device, we should generalise this such that it can be loaded on `rocm` or `mps` without having to be explicit.

Open a checkpoint with inference on any non cuda

### Version

0.7.0

...

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ECMWF

HCookie:
## Description
Lazily resolve the device to be used, 
Removes the default 'cuda', and instead checks for device availability

Simplifies loading and running on non-cuda systems

Closes #296 


src/anemoi/inference/device.py
```diff
@@ -0,0 +1,63 @@
...
+                self._device = "cpu"
+                break
+
+        if self._device is None:
+            raise RuntimeError(f"No supported devices found among: {self.prefer}")
+
+        return self._device
+
+    def __str__(self):
+        return self.resolve()
```
tmi, line 45: this makes it serialization-time evaluated, whereas we want deserialization-time evaluated, no?

tmi: unless I misunderstood something, I would have expected something like `_device: str|None` and a `@property device` which returns `self._device` if not None, otherwise does the availability check...

tmi (dismissed):

src/anemoi/inference/runner.py
```diff
@@ -178,6 +180,12 @@ def checkpoint(self) -> Checkpoint:
         """
         return self._checkpoint
 
+    @property
+    def device(self) -> str:
+        if self._device is None:
+            self._device = str(get_available_device())
```
aaron-hopkinson, line 186: Is it really necessary to cast the result to a string here? Could you not just use the `torch.device`?

HCookie: I could, but then the type between the default and the config would be different.

HCookie: But, saying that I could force the string to a `torch.device`?

src/anemoi/inference/runner.py
```diff
@@ -624,7 +639,7 @@ def forecast(
 
             # Predict next state of atmosphere
             with (
-                torch.autocast(device_type=self.device, dtype=self.autocast),
+                torch.autocast(device_type=str(self.device), dtype=self.autocast),
```
gmertes, line 642: Is the str cast still needed here if it is already a torch.device ?

gmertes (approved):

src/anemoi/inference/runner.py
HCookie: Yeah, 100% autocast only works with strings oddly

PR #299, opened 2025-08-04
build(pytest): configure pytest to remove all tmpdirs

VeraChristina:
## Description
Disable tmpdir retention in pytest. By default, pytest keeps the last 3 tmpdirs when using the tmp_path fixture.

We won't leave tmpdirs on runners when running pytest in the ci

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,6 +105,8 @@ markers = [
   "skip_on_hpc: mark a test that should not be run on HPC",
 ]
 testpaths = "tests"
+tmp_path_retention_policy = "none"
+tmp_path_retention_count = 0
 
 [tool.mypy]
 exclude = [ "docs/" ]


aaron-hopkinson:
@VeraChristina do you plan to roll this out across the other repos?

VeraChristina:
> @VeraChristina do you plan to roll this out across the other repos?

It's already set in training, I haven't checked datasets and transform yet. Thoughts? (The other option is to not change the pytest setting and just disable it in the ci)

aaron-hopkinson:
@VeraChristina 
> The other option is to not change the pytest setting and just disable it in the ci)

That would probably be my preference. Having old data lying around by default is (as far as I know) by design (so you can look at failing test data). I appreciate this behaviour isn't useful in the CI where the data isn't going to be looked at though....

VeraChristina:
> That would probably be my preference. Having old data lying around by default is (as far as I know) by design (so you can look at failing test data). I appreciate this behaviour isn't useful in the CI where the data isn't going to be looked at though....

Makes sense. (And if exceeding temp dir size in training (where this is currently set to none) causes too much hassle for people, we could set it to only keep the last folder instead of three.) I'll wait to hear what @gmertes thinks

Issue #273, opened 2025-07-09
Add more flavour models


gmertes:
To begin with: 

- [ ] ensemble. Doesn't have to be a different model, it's just setting different grib keys at the output (stream, type and number).
- [x] LAM
- [x] stretched grid

PR #308, opened 2025-08-11
test: Improve integration test and add RMI LAM

### LINKED ISSUE #272: Helper script to add new models
Starting from an o48 checkpoint file trained with anemoi-models, have a script that:
- extracts and sanitises the metadata json
- extract supporting arrays
- get some test data from mars
- upload the input grib and supporting arrays to S3
- initialise the test config and put metadata json in the integration test folder (to be committed)

gmertes:
## Description
This PR adds the RMI LAM model to the integration test. As part of this, the following was done:

1. Add a helper script to add new models.
2. The mock model now passes prognostic input variables of the last input time step to the output. Diagnostics are output as all ones. This allows us to test the full data pipeline and check output values against the input.
3. Add a new `${s3:}` resolver available in the test config, this resolves to the model path on S3 where test data is located.
4. The integration test now downloads supporting arrays from S3 (if they are present in the metadata). Sequential downloading was slower than I liked (the download itself is fast, but opening the connection takes a second), using a few threads speeds it up.
5. Add integration test docs
6. Small fix in the unit tests: rename TestingProcessor to DummyProcessor to avoid pytest trying to pick it up as a test class, and the simple model was missing prognostics in the input data indices.
7. Add support for multiple input files. Not used in this model but will be used by MeteoSwiss.

### Helper script
The helper script is run on a real inference checkpoint, trained at a reduced resolution. See the docs for more explanation.
```shell
$ python anemoi-inference/tests/integration/add_new_model.py -h

usage: add_new_model.py [-h] [--files [FILES ...]] [--overwrite] model checkpoint

Add a new model for integration tests.
Running this script will create a new model directory in tests/integration
with metadata and config files, and upload necessary files to S3.

positional arguments:
  model             Model name. Can only contain alphanumeric, underscores, hyphens, or dots
  checkpoint        Path to the inference checkpoint file

options:
  -h, --help        show this help message and exit
  --files [FILES ...], -f [FILES ...]
                    Additional files to upload to the model directory on S3
  --overwrite, -o   Overwrite existing files
```

It extracts the metadata from the checkpoint and stores it in `tests/integration/{model}/` together with an example config. Supporting arrays are also extracted and uploaded to S3 under `s3://ml-tests/test-data/samples/anemoi-integration-tests/inference/{model}/`. 

I did not put this script in the CLI because this is a developer-only script, and will create files to its relative location in the repo.

### RMI LAM
The model was trained on a O48 inner and O32 boundary region. Reduced testing datasets were made available in the catalogue. Because the RMI setup runs from a zarr dataset, a further-reduced 5 day version of the test datasets was made locally to be used as input data. These were zipped and provided to the above script as extra `--files` and uploaded to S3. 

The test configuration uses the facility of anemoi-datasets to download zarr zips at runtime. Because the input datasets are tiny (~10MB), this works well and is fast. 

Note: when doing this, care needs to be taken that there is no subdirectory in the zip: the root path of the zarr should be the root path of the zip. This was done with: `(cd aifs-ea-an-oper-0001-mars-o48-2020-6h.zarr && zip -r ../aifs-ea-an-oper-0001-mars-o48-2020-6h.zarr.zip .)`

Because the mock integration test model now passes prognostic inputs to the output, the new `check_lam_with_xarray` check compares the values in the output file to a reference dataset (the input). The test config is self-explanatory: we pass the same dataset used for the inner region to the check, which in turn compares the value of each prognostic variable. We also check that the number of grid points matches the mask in the checkpoint.

closes #272
#273

----
📚 Documentation preview 📚: https://anemoi-inference--308.org.readthedocs.build/en/308/

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -80,18 +80,19 @@ def test_setup(request, get_test_data: GetTestData, tmp_path: Path) -> Setup:
     with open(INTEGRATION_ROOT / model / "metadata.json") as f:
         metadata = json.load(f)
 
-    if supporting_arrays := metadata.get("supporting_arrays_paths", {}).keys():
+    supporting_arrays = {}
+    if array_names := list(metadata.get("supporting_arrays_paths", {})):
 
-        def load_array(array_name):
-            return array_name, np.load(get_test_data(f"{s3_path}/supporting-arrays/{array_name}.npy"))
+        def load_array(name):
+            return name, np.load(get_test_data(f"{s3_path}/supporting-arrays/{name}.npy"))
 
         with ThreadPoolExecutor(max_workers=4) as executor:
-            results = list(executor.map(load_array, list(supporting_arrays)))
+            results = list(executor.map(load_array, array_names))
 
         supporting_arrays = dict(results)
         LOG.info(f"Supporting arrays: {supporting_arrays.keys()}")
 
-    save_fake_checkpoint(metadata, checkpoint_path, supporting_arrays=supporting_arrays or {})
+    save_fake_checkpoint(metadata, checkpoint_path, supporting_arrays=supporting_arrays)
 
     # substitute inference config with real paths
     OmegaConf.register_new_resolver("input", lambda i=0: str(input_data[i]), replace=True)


src/anemoi/inference/testing/checks.py
```diff
@@ -114,3 +115,46 @@ def check_with_xarray(
...
+    ds = xr.open_dataset(file)
+
+    # check shape of inner region against the mask in the checkpoint
+    mask = checkpoint.load_supporting_array(f"{mask}/cutout_mask")
+    for var in ds.data_vars:
+        assert ds[var].shape[-1] == np.sum(
+            mask
+        ), f"Variable {var} shape {ds[var].shape[-1]} does not match mask size {np.sum(mask)}."
+
+    # check that the extracted inner region matches the reference input dataset
```
theissenhelen, line 144: Does anemoi-inference also do the region check?

gmertes: No, at inference time it just applies the cutout mask from the checkpoint to the output and that is written to disk. 
There's also no ground-truth at inference time to do any checks against, we only have it here because with the mock model the ground truth is the input.

tests/integration/test_integration.py
```diff
@@ -53,21 +58,46 @@ def test_setup(request, get_test_data: GetTestData, tmp_path: Path) -> Setup:
...
+    os.chdir(workdir)
 
     # prepare checkpoint
     checkpoint_path = tmp_path / Path("checkpoint.ckpt")
     with open(INTEGRATION_ROOT / model / "metadata.json") as f:
         metadata = json.load(f)
-    # TODO: also get supporting arrays from S3 if needed and store them in the checkpoint
-    save_fake_checkpoint(metadata, checkpoint_path)
+
+    if supporting_arrays := metadata.get("supporting_arrays_paths", {}).keys():
```
theissenhelen, line 83:
```suggestion
    if supporting_arrays_keys := metadata.get("supporting_arrays_paths", {}).keys():
```

tests/integration/test_integration.py
```diff
@@ -53,21 +58,46 @@ def test_setup(request, get_test_data: GetTestData, tmp_path: Path) -> Setup:
...
+
+        def load_array(array_name):
+            return array_name, np.load(get_test_data(f"{s3_path}/supporting-arrays/{array_name}.npy"))
+
+        with ThreadPoolExecutor(max_workers=4) as executor:
+            results = list(executor.map(load_array, list(supporting_arrays)))
```
theissenhelen, line 89:
```suggestion
            results = list(executor.map(load_array, supporting_arrays_keys))
```

tests/integration/test_integration.py
```diff
@@ -53,21 +58,46 @@ def test_setup(request, get_test_data: GetTestData, tmp_path: Path) -> Setup:
...
+
+        supporting_arrays = dict(results)
+        LOG.info(f"Supporting arrays: {supporting_arrays.keys()}")
+
+    save_fake_checkpoint(metadata, checkpoint_path, supporting_arrays=supporting_arrays or {})
```
theissenhelen, line 94:
```suggestion
    save_fake_checkpoint(metadata, checkpoint_path, supporting_arrays=supporting_arrays_keys or {})
```

PR #328, opened 2025-09-19
fix: Inference with single variable

nilsmkMET:
## Description
Fix for running inference with single variable output

When running anemoi-inference, and only predicting one single variable, the run will fail due to the "output"-variable being squeezed into a 1D array. Anemoi-infernece expect it to be 2D.
The same problem occurs when running the downscaling runner (as noted by user @paulina-t )

N/A

##  Additional notes ##
N/A

### DIFF
--- a/src/anemoi/inference/runner.py
+++ b/src/anemoi/inference/runner.py
@@ -646,10 +646,7 @@ class Runner(Context):
                 y_pred = self.predict_step(self.model, input_tensor_torch, fcstep=s, step=step, date=date)
 
             output = torch.squeeze(y_pred, dim=(0, 1))  # shape: (values, variables)
-            if output.ndim == 1:
-                # If there is only one variable, we need to add a dimension
-                output = output.reshape(output.shape[0], 1)  # shape: (values, variables)
-
+            
             # Update state
             with ProfilingLabel("Updating state (CPU)", self.use_profiler):
                 for i in range(output.shape[1]):
--- a/src/anemoi/inference/runner.py
+++ b/src/anemoi/inference/runner.py
@@ -645,7 +645,7 @@ class Runner(Context):
             ):
                 y_pred = self.predict_step(self.model, input_tensor_torch, fcstep=s, step=step, date=date)
 
-            output = torch.squeeze(y_pred)  # shape: (values, variables)
+            output = torch.squeeze(y_pred, dim=(0,1))  # shape: (values, variables)
             if output.ndim == 1:
                 # If there is only one variable, we need to add a dimension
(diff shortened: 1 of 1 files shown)


src/anemoi/inference/runner.py
```diff
@@ -646,6 +646,9 @@ def forecast(
                 y_pred = self.predict_step(self.model, input_tensor_torch, fcstep=s, step=step, date=date)
 
             output = torch.squeeze(y_pred)  # shape: (values, variables)
```
gmertes, line 648:
```suggestion
            output = torch.squeeze(y_pred, dim=(0,1))  # shape: (values, variables)
```

gmertes (commented):
Thanks for the fix! This already looks okay to me, but instead of adding the dimension back in after the squeeze, it would be simpler to just use the `dim` option of torch.squeeze. This takes a tuple, so I believe `torch.squeeze(y_pred, dim=(0,1))` should do the trick).

paulina-t:
Thanks @nilsmkMET and @gmertes.

This issue also appears in the feat/downscaling branch.

I fixed it by adding again the dimension in `predict_step` in `downscaling.py`

```
    def predict_step(self, model, input_tensor_torch, date, step, **kwargs):
        input_date = date - step
        low_res_tensor = input_tensor_torch
        high_res_tensor = self._prepare_high_res_input_tensor(input_date)

        LOG.info("Low res tensor shape: %s", low_res_tensor.shape)
        LOG.info("High res tensor shape: %s", high_res_tensor.shape)

        extra_args = self.extra_config.get("extra_args", {})

        residual_output_tensor = model.predict_step(low_res_tensor, high_res_tensor, extra_args=extra_args, **kwargs)
        residual_output_numpy = np.squeeze(residual_output_tensor.cpu().numpy())
        if residual_output_numpy.ndim == 1:
            residual_output_numpy = residual_output_numpy[:, np.newaxis]
            
            
```
         
But I'm happy with your fix, @gmertes. Could you add it there too, so I don't need to open another PR?

src/anemoi/inference/runner.py
nilsmkMET: Yes, seems like a more elegant solution, thanks

nilsmkMET:
Yes, sorry forgot that. Looks good now?

paulina-t:
Hi again, 

I just tested this fix with the downscaling branch, but it didn't work. Did it work for you, @nilsmkMET?

Moreover, as suggested by @JoffreyDumontLeBrazidec, I'll push a new branch created from `feat/downscaling)` used in the C3S2-36b1 project. This is still work in progress, hope that's fine with you @gmertes

PR #331, opened 2025-09-24
feat: replace no grad with inference mode

tmi:
replaces torch.no_grad with torch.inference_mode

1/ should gain even more speedup as version tracking / counters, cf https://docs.pytorch.org/docs/2.8/generated/torch.autograd.grad_mode.inference_mode.html . Note this additionally disables AD in forward mode, but I guess this is not being used anywhere
2/ replacet the set call with context managers, to not pollute the global state

### DIFF
--- a/src/anemoi/inference/runner.py
+++ b/src/anemoi/inference/runner.py
@@ -594,112 +594,118 @@ class Runner(Context):
         Any
             The forecasted state.
         """
-        self.model.eval()
+        # NOTE we are not using decorator of the top level function as we anticipate lazy torch load
+        with torch.inference_mode():
+            self.model.eval()
 
-        torch.set_grad_enabled(False)
+            # Create pytorch input tensor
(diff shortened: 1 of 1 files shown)


gmertes (dismissed):
Thanks for this. Like you say this should only have a positive impact on performance and no impact on the model itself, but just for paranoia sake I'm going to do a verification with AIFS first to make sure.

(ignore the requested changes, that's just to block the PR from being merged by someone else)

tmi:
understood, I've switched to context manager -- I mean, we could write a decorator lazifier, its neither overly hard nor ugly. Esp since there will be already a custom lazy loader in the codebase. But I don't really like doing that -- prevents type checking and makes code bit less readable. And the alternative is not terrible

but if you run into this more often, it may make sense ofc. Together with LazyGlobalConstantEvaluator, LazyTypeEvaluator, LazyParentClassEvaluator... lazy importing in python aint for faint hearted :)

PR #333, opened 2025-09-25
fix(parallel runner): Support base class with main_argument decorator

### LINKED ISSUE #309: Parallel runner fails when base runner has main_argument decorator
If a runner has the main_argument decorator, the object stored in the registry is a function (produced by the decorator) and not the original class. If such a runner is used as the base_runner for the parallel runner, it fails because the original class is lost. 

The parallel runner expects the registry to contain the class so that it can use it as the base class for the ParallelRunner.

Thanks @mtgarciag for finding the bug.

gmertes:
## Description
If the `main_argument` decorator is used on a class and subsequently stored in the registry, the registered object becomes a function. In most cases this is not a problem, except in the parallel runner which needs the original class of the base runner in the registry, as described in #309.

Refactor this so the wrapping is done with a class instead of a function.

The decorator is not used on functions anywhere, and I don't think there is a use case to do so. This removes support for the decorator to be used on functions, and only support classes. 

fixes #309 

### DIFF
--- a/src/anemoi/inference/decorators.py
+++ b/src/anemoi/inference/decorators.py
@@ -18,7 +18,7 @@ F = TypeVar("F", bound=type)
 
 
 class main_argument:
-    """Decorator to set the main argument of a class.
+    """Decorator to set the main argument of a class. Only for classes with a 'context' argument.
 
     For example:
     ```


src/anemoi/inference/decorators.py
```diff
@@ -50,43 +51,16 @@ def __init__(self, name: str):
...
 
-            Returns
-            -------
-            Any
-                The result of the decorated function.
-            """
-            if main is not MARKER:
-                kwargs[self.name] = main
+        if not isinstance(cls, type):
+            raise TypeError("main_argument can only be used to decorate classes")
```
HCookie, line 58:
```suggestion
            raise TypeError("'main_argument' can only be used to decorate classes")
```
Semantics, so happy to ignore

PR #339, opened 2025-10-03
fix: Consolidate path behaviour on outputs

HCookie:
## Description

- Make dir when none existed
- Use a single templating function
- Clean up typing and function signatures
- Add further tests

##  Additional notes ##

### DIFF
--- a/src/anemoi/inference/outputs/netcdf.py
+++ b/src/anemoi/inference/outputs/netcdf.py
@@ -10,6 +10,7 @@
 import logging
 import os
 import threading
+from pathlib import Path
 
 import numpy as np
 
@@ -36,7 +37,7 @@ class NetCDFOutput(Output):
     def __init__(
         self,
         context: Context,
-        path: str,
+        path: Path,
         variables: list[str] | None = None,
         post_processors: list[ProcessorConfig] | None = None,
         output_frequency: int | None = None,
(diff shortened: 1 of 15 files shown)


tests/integration/single-o48-1.1/config.yaml
```diff
@@ -83,3 +83,132 @@
...
+      grib: ${input:}
+    output:
+      raw:
+        dir: ${output:}
+        template: "{date}_step{step}.npz"
+
+
+- name: grib-in-tee-output
+  input: input.grib
+  output: output.grib
```
gmertes, line 199:
```suggestion
  output: output
```

To indicate that it's a folder, not a file

src/anemoi/inference/outputs/netcdf.py
```diff
@@ -36,22 +37,23 @@ class NetCDFOutput(Output):
     def __init__(
         self,
         context: Context,
-        path: str,
+        path: Path,
```
gmertes, line 40: Technically it's still a string when it's instantiated from a config, I guess the type hint should be `str | Path` ?

There's also a little duplication going on with the casting to `Path` and creating the parent directories in all the different outputs. Would a new decorator be a good idea, something like `@ensure_path("path")` ?

But also happy to leave it like this.

PR #347, opened 2025-10-15
test(integration): Add meteoswiss-sgm-cosmo

gmertes:
## Description
Add the MeteoSwiss stretched-grid cosmo model to the integration test. 

o48 global and 8km local over a tiny region. Global test data coming from IFS, local from MeteoSwiss. 

Requires eccodes cosmo local definitions, and custom grib templates for the local region. 

TODO: 
- [x] test isolation to forcefully reload eccodes local definitions
- [x] checks
- [x] docs 

The cosmo model runs in its own workflow due to conflicting eccodes local definition. When `pytest` is invoked, it will skip the cosmo test by default. A new separate workflow calls `pytest --cosmo`, and this only runs cosmo. 

#273

----
📚 Documentation preview 📚: https://anemoi-inference--347.org.readthedocs.build/en/347/

### DIFF
--- a/src/anemoi/inference/testing/checks.py
+++ b/src/anemoi/inference/testing/checks.py
@@ -100,7 +100,7 @@ def check_grib_cutout(
     if not reference_datetime:
         reference_datetime = ref_ds.order_by(valid_datetime="ascending")[-1].metadata("valid_time")
 
-    assert len(ds) > 0, "No fields found in the LAM GRIB file."
+    assert len(ds) > 0, "No fields found in the output GRIB file."
     assert len(ref_ds) > 0, "No fields found in the reference GRIB file."
 
     mask = checkpoint.load_supporting_array(f"{mask}/cutout_mask")
@@ -109,13 +109,13 @@ def check_grib_cutout(
     ]
 
     for param in prognostic_params:
-        lam_fields = ds.sel(param=param)
+        fields = ds.sel(param=param)
         ref_fields = ref_ds.sel(param=param, valid_time=reference_datetime)
 
-        assert len(lam_fields) > 0, f"No fields found for variable {param} in LAM file."
+        assert len(fields) > 0, f"No fields found for variable {param} in output file."
         assert len(ref_fields) > 0, f"No fields found for variable {param} in reference file at {reference_datetime}."
 
-        for field in lam_fields:
+        for field in fields:
             assert field.values.shape[-1] == np.sum(
                 mask
             ), f"Variable {param} shape {field.shape[-1]} does not match mask size {np.sum(mask)}."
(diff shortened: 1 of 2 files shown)


tests/integration/test_integration.py
```diff
@@ -113,6 +128,16 @@ def load_array(name):
...
 
+    if test_setup.config.get("cosmo"):
+        import sys
+        from pathlib import Path
+
+        import eccodes
+
+        cosmo = str(Path(sys.prefix) / "share/eccodes-cosmo-resources/definitions")
+        current_path = eccodes.codes_definition_path()
+        eccodes.codes_set_definitions_path(f"{cosmo}:{current_path}")
```
HCookie: Setting some sort of env var for use in the config with `env` will likely be more scalable.

gmertes:
PR is now ready for review. The CI is being flaky due to network issues on the CCI (the data download is failing), but all the tests are actually passing.

tests/integration/test_integration.py
```diff
@@ -32,6 +34,15 @@
 
 INTEGRATION_ROOT = Path(__file__).resolve().parent
 
+
+def _marker(model: str, config: DictConfig):
+    """Add markers at collection time based on the model test config."""
+    marks = []
+    if "cosmo" in model or config.get("cosmo"):
+        marks.append(pytest.mark.cosmo)
+    return marks
```
frazane: Not a big fan of this one. I would prefer if we had an explicit `markers` entry in the configs.

tests/integration/meteoswiss-sgm-cosmo/config.yaml
```diff
@@ -0,0 +1,69 @@
...
+    post_processors:
+      - extract_from_state: lam_0
+    write_initial_state: false
+    input:
+      cutout:
+        lam_0:
+          grib:
+            path: ${input:0}
+            namer:
+              rules:
```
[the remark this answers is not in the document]
gmertes: Yes it's a good idea, although I would suggest we do that in a follow up?

tests/integration/test_integration.py
gmertes: Good idea, I changed it in c0a5b772c2b0e2568c7e9a9247046bb583d0b5a1

src/anemoi/inference/testing/checks.py
```diff
@@ -78,6 +78,52 @@ def check_grib(
...
+@testing_registry.register("check_grib_lam")
+def check_grib_lam(
+    *,
+    file: Path,
+    checkpoint: "Checkpoint",
+    reference_grib: str,
+    reference_datetime: str | None = None,
+    mask="lam_0",
+    **kwargs,
+):
```
frazane: Nit: is the cutout operation specific to the stretched grid model approach? If so, it could better to rename this `check_grib_sgm` (Stretched Grid Model). If it's common to both LAM and SGM, then maybe rename this to `check_grib_cutout`? Same for the function below.

gmertes: Good point, yes these checks are shared between SGM and LAM cutout. I renamed them in 3d33ee0410c36d6501d750e2f8854cb3a29c870e

frazane (approved):
Looks good to me, thank you! Looking forward to the next one :-)

PR #355, opened 2025-10-17
fix(input cutout): `CutoutInput` improvements and upgrades

### LINKED ISSUE #356: Cutout input sources should be specified as a list
### What happened?

Currently, we provide a dictionary of sources to the cutout input. However, this is risky because the cutout operation to combine sources is order-dependent, and dictionaries do not enforce any order. Whenever one updates a config by opening it in python, making some overrides, and writing back, the order is not preserved. I realize one can override from the command line, but in some workflows one might prefer to generate a new config for better traceability. 

There are measures to take to preserve the order that users might take, but I find that an easier option would be to just specify the cutout input as a list of single-entry dictionaries, such as 

```yaml
input:
  cutout:
    - lam: ...
    - global: ...
```

Start from a default config with a cutout input such as:

```yaml
input:
  cutout:
    lam: ...
    global: ...
```

then make some unrelated changes to it, such as updating the date. When writing back the updated config, the order will have changed to 

```yaml
input:
  cutout:
    global: ...
    lam: ...
```

which will make the inference fail.

### Version

v0.8.0

x86 Linus

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

MeteoSwiss

HCookie:
## Description

A couple improvements to the cutout input:
- Remove `pre_processors` from top level cutout
- Add `_input` to cutout states
- Alter `CutoutInput` to take a list to provide an explicit ordering

Closes #356 

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-inference--355.org.readthedocs.build/en/355/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # anemoi-inference
 
-**DISCLAIMER**
-This project is **BETA** and will be **Experimental** for the foreseeable future.
-Interfaces and functionality are likely to change, and the project itself may be scrapped.
-**DO NOT** use this software in any project/software that is operational.
+[![Static Badge](https://github.com/ecmwf/codex/raw/refs/heads/main/Project%20Maturity/incubating_badge.svg)](https://github.com/ecmwf/codex/raw/refs/heads/main/Project%20Maturity#incubating)
 
+
+> \[!IMPORTANT\]
+> This software is **Incubating** and subject to ECMWF's guidelines on [Software Maturity](https://github.com/ecmwf/codex/raw/refs/heads/main/Project%20Maturity).
 
 
 ## Documentation


HCookie:
Sure, I'm happy to change it, 
I think it will likely end up looking like
```yaml
input:
  cutout:
		inputs:
	    - lam: ...
	    - global: ...
```

src/anemoi/inference/decorators.py
```diff
@@ -43,17 +43,29 @@ def __init__(context, encoding=None, path=None, archive_requests=None):
...
+    @main_argument("paths", capture_all_args=True)
+    class MultiFileOutput
+        def __init__(context, paths=None, archive_requests=None):
+            ...
+    output = MultiFileOutput(context, "out1.grib", "out2.grib")
+    ```
     """
 
-    def __init__(self, name: str):
+    def __init__(self, name: str, *, capture_all_args: bool = False):
```
gmertes: Can you explain the usefulness of this new flag? Not sure about the benefit of this extra complexity. The decorator should really only be used in combination with the config factory, but I don't see how it's useful in a config setting and in relation to the cutout input. 

Doesn't this already work without the new flag?

```
output: 
  grib: ["out1.grib", "out2.grib"]
```

gmertes (commented):
Is this a breaking change (new config format)?

PR #365, opened 2025-10-28
feat(parallel runner)!: Add cluster environments

HCookie:
## Description
Add cluster environments to delegate concern on parallel inference.

Backwards compatible with slurm running, as automatic detection of the cluster is implemented,

Allows for easier extension to additional clusters, with support now already for slurm, mpi, torchrun, distributed and manual invocation.

## Manual Cluster
```yaml
runner:
   parallel:
       cluster:
           manual: 4
```

## Automatic
```yaml
runner: parallel
```

## Tests

- [x] torchrun
- [x] slurm
- [x] mpi
- [x] manual
- [x] AzureML

Extends parallel running to allow for easy implementation of exotic clusters

----
📚 Documentation preview 📚: https://anemoi-inference--365.org.readthedocs.build/en/365/


src/anemoi/inference/clusters/cluster.py
```diff
@@ -0,0 +1,183 @@
...
+from anemoi.inference.lazy import torch
+
+ADDRESS = NamedTuple("Address", [("host", str), ("port", int)])
+LOG = logging.getLogger(__name__)
+
+
+class Cluster(ABC):
+    """Abstract base class for cluster and parallel environment handling."""
+
+    _model_comm_group: torch.distributed.ProcessGroup | None = None  # type: ignore
```
tmi: one thought here: I think torch.distributed will be eventually superseded by monarch. But then that may take year+

src/anemoi/inference/clusters/cluster.py
```diff
@@ -0,0 +1,183 @@
...
+        _ = kwargs  # To avoid unused variable warning
+
+    @classmethod
+    @abstractmethod
+    def used(cls) -> bool:
+        """Check if this cluster is valid in the current environment."""
+        pass
+
+    def spawn(self, fn: Any, *args: Any) -> None:
+        """Spawn processes in the cluster environment.
```
tmi: To check I understand the intention. Let's say we want a Cascade-based cluster, which already has gang capability. This means we would launch a gang, and each of the tasks would get world size and rank and coordinator address

what is each task's callable then, from the PoV of this class? I would expect something that calls the constructor of this class to populate the master address/port and world size and rank from the envvars, and call initialise. And then continue with the actual torch code

but is there a place for spawn then? Is spawn always expected to be invoked, or is it only to be so for "manual" clusters?

src/anemoi/inference/clusters/cluster.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Parameters
+        ----------
+        context : Context
+            Runner context
+        kwargs : Any
+            Additional keyword arguments
+        """
+        self.context = context
+        _ = kwargs  # To avoid unused variable warning
```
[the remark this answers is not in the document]
tmi: in the manual case, you dont propagate the kwargs to the super (correctly!)

docs/inference/parallel.rst
```diff
@@ -110,3 +203,149 @@ job across 4 GPUs with SLURM.
...
+.. code:: bash
+
+   #!/bin/bash
+
+   source /path/to/venv/bin/activate
+
+   torchrun --nproc_per_node=4 \
+            --nnodes=1 \
+            --node_rank=0 \
+            --master_addr=localhost \
```
cathalobrien: Nice example, this will only work for a single node case right?

docs/inference/parallel.rst
```diff
@@ -110,3 +203,149 @@ job across 4 GPUs with SLURM.
...
+   #SBATCH --ntasks-per-node=4
+   #SBATCH --gpus-per-node=4
+   #SBATCH --cpus-per-task=8
+   #SBATCH --time=0:05:00
+   #SBATCH --output=outputs/parallel_inf_mpi.%j.out
+
+   source /path/to/venv/bin/activate
+
+   # Set master address and port for communication
+   export MASTER_ADDR=$(hostname)
```
cathalobrien: Will this work for the multinode case? Since your example is a slurm script we could use this instead?

```
MASTER_ADDR=$(scontrol show hostname $SLURM_NODELIST | head -n 1)
export MASTER_ADDR=$(nslookup $MASTER_ADDR | grep -oP '(?<=Address: ).*')
```

docs/inference/parallel.rst
```diff
@@ -2,52 +2,142 @@
...
+      -  ``RANK``, ``LOCAL_RANK``, ``WORLD_SIZE``, ``MASTER_ADDR``,
+         ``MASTER_PORT``
+
+Manual Cluster Configuration
+============================
+
+If you are running in an environment without automatic detection, use
+the manual cluster
+(:class:`anemoi.inference.clusters.manual.ManualCluster`) by specifying
+the ``world_size`` (number of GPUs):
```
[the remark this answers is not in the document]
HCookie: A bit of both, but I can clarify

docs/inference/parallel.rst
HCookie: Yeah for now...

docs/inference/parallel.rst
```diff
@@ -110,3 +203,149 @@ job across 4 GPUs with SLURM.
...
+      -  Check firewall settings and ensure all nodes can communicate.
+         Verify ``MASTER_ADDR`` is accessible from all ranks.
+
+Verification Checklist
+======================
+
+Before running parallel inference, verify:
+
+#. ✓ Anemoi-Models version >= v0.4.2
+#. ✓ Multiple GPUs available (``nvidia-smi`` or equivalent)
```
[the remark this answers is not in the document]
HCookie: Addressed in [2ceeca1](https://github.com/ecmwf/anemoi-inference/pull/365/commits/2ceeca11fb26d129112b4f4f442a7d635db427d2)

docs/inference/parallel.rst
```diff
@@ -110,3 +203,149 @@ job across 4 GPUs with SLURM.
...
+
+   # Set master address and port for communication
+   export MASTER_ADDR=$(hostname)
+   export MASTER_PORT=29500
+
+   mpirun -np 4 anemoi-inference run parallel.yaml
+
+.. note::
+
+   You can optionally use the ``mpi`` torch backend by configuring:
```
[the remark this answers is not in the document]
HCookie: Addressed in [2ceeca1](https://github.com/ecmwf/anemoi-inference/pull/365/commits/2ceeca11fb26d129112b4f4f442a7d635db427d2)

src/anemoi/inference/config/run.py
```diff
@@ -32,7 +32,7 @@ class RunConfiguration(Configuration):
     """A path to an Anemoi checkpoint file."""
 
     runner: str | dict[str, Any] = "default"
-    """The runner to use."""
+    """The runner to use. If using `parallel`, can set `cluster` options here"""
```
gmertes:
```suggestion
    """The runner to use. For runners that take extra options, they can be passed here in dictionary format with the runner name as the top level dictionary key. For example, if using `parallel`, the `cluster` option can be set here."""
```

gmertes (approved):
🚀

PR #374, opened 2025-11-05
fix: OmegaConf loading differs from yaml

### LINKED ISSUE #343: Some valid anemoi-dataset configs are no longer accepted.
### What happened?

Consider the (partial) config
```yaml
input:
  dataset:
    cutout:
      - dataset: bla
      - dataset:
        dataset:
        - dataset:
          dataset: bli
          start: 2020-10-01T18:00:00
          end: 2020-10-02T00:00:00
        - dataset:
          dataset: blu
          start: 2020-10-02T06:00:00
          end: 2020-10-09T00:00:00
        reorder: ['all_kind_of_variables']
    adjust: all
```
This works perfectly in anemoi-inference==0.6.2 but fails in the latest anemoi-inference release. (See log below). 

Note that it is not an issue with the yaml, as 
```python
with open("above.yaml", 'r') as f:
    config = yaml.safe_load(f)
```
leads to a nice dictionary. [That furthermore is accepted by anemoi-datasets, irrespective of version]

Looks like this is OmegaConf related and hence probably introduced in 0.6.3.

Run inference with a config that has input of the type above.

### Version

0.7.3

irrelevant

### Relevant log output

```shell
File ".../lib/python3.12/site-packages/omegaconf/_utils.py", line 144, in construct_mapping
    raise yaml.constructor.ConstructorError(


yaml.constructor.ConstructorError: while constructing a mapping
  in "XXX.yaml", line Y, column 9
found duplicate key dataset
  in "XXX.yaml", line Y+1, column 9
```

### Accompanying data

_No response_

### Organisation

RMI

HCookie:
## Description
Load with yaml, and dump before omegaconf

Allows for valid yaml to be loaded with omegaconf

Closes #343 

##  Additional notes ##


dietervdb-meteo (changes requested):
LGTM and I tested that it indeed solves #343 

Small comment since we now `import yaml` shouldn't this be listed in the dependencies in the `pyproject.toml`? or is this dependency already fullfilled through some other package? If not, looks like we already have a dependency on `pyyaml` would that be enough?

gmertes (commented):
I don't fully understand why omegaconf is unable to parse this yaml. The fix seems fine to me, if a bit clunky. Can you add a comment with link to the issue so our future selves still know why the clunkyness is there.

PR #383, opened 2025-11-17
feat(grib): Improve template manager

gmertes:
## Description
Improve the grib template manager in a backwards compatible way.

- [x] Improve documentation on how users can provide their own templates
- [x]  For the `samples` provider, add the option for the sample index to be specified directly in the inference config
- [x]  For the `file` provider, add an option for the file to contain many parameters, and do a lookup based on the param name. 
- [x]  In the case of grib input, the template manager automatically injects templates found in the input grib at the highest priority. Refactor this so that this becomes its own standalone template provider called `input`, this gives more flexibility to enable/disable combinations of template providers and set the priority..
- [x] `input` provider fallback mode: force an output variable to use one of the input variables
- [x]  Improve logging of the template manager. Right now it only logs in the case of errors. Also log in the case of success so it's easier to see which templates are being used from which provider. 
- [x] tests for everything, including backwards compatibility

----
📚 Documentation preview 📚: https://anemoi-inference--383.org.readthedocs.build/en/383/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
   - id: python-check-blanket-noqa # Check for # noqa: all
   - id: python-no-log-warn # Check for log.warn
 - repo: https://github.com/psf/black-pre-commit-mirror
-  rev: 25.9.0
+  rev: 25.11.0
   hooks:
   - id: black
     args: [--line-length=120]
@@ -41,7 +41,7 @@ repos:
     - --profile black
     - --project anemoi
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.3
+  rev: v0.14.7
   hooks:
   - id: ruff
     args:
@@ -50,7 +50,7 @@ repos:
     - --exit-non-zero-on-fix
     - --exclude=docs/**/*_.py
 - repo: https://github.com/sphinx-contrib/sphinx-lint
-  rev: v1.0.1
+  rev: v1.0.2
   hooks:
   - id: sphinx-lint
 # For now, we use it. But it does not support a lot of sphinx features
@@ -65,7 +65,7 @@ repos:
   - id: docconvert
     args: ["numpy"]
 - repo: https://github.com/tox-dev/pyproject-fmt
-  rev: "v2.11.0"
+  rev: "v2.11.1"
   hooks:
   - id: pyproject-fmt
     args: ["--max-supported-python", "3.12"]


src/anemoi/inference/grib/templates/__init__.py
```diff
@@ -54,7 +54,10 @@ def __init__(self, manager: Any) -> None:
         """
         self.manager = manager
 
-    def template(self, variable: str, lookup: dict[str, Any]) -> ekd.Field:
+    def __repr__(self):
+        return f"{self.__class__.__name__}"
```
frazane: It could be helpful to see the kwargs in the string representation (assuming it will be used in logging somewhere?).

src/anemoi/inference/grib/templates/__init__.py
```diff
@@ -75,21 +80,24 @@ def template(self, variable: str, lookup: dict[str, Any]) -> ekd.Field:
 class IndexTemplateProvider(TemplateProvider):
     """Template provider based on an index file."""
 
-    def __init__(self, manager: Any, index_path: str) -> None:
+    def __init__(self, manager: Any, index_path: str | list) -> None:
```
frazane: Would it be possible to replace `Any` with something else? A base class? Protocol?

src/anemoi/inference/grib/templates/__init__.py
```diff
@@ -75,21 +80,24 @@ def template(self, variable: str, lookup: dict[str, Any]) -> ekd.Field:
...
-        with open(index_path) as f:
-            self.templates = yaml.safe_load(f)
+        if isinstance(index_path, str):
+            with open(index_path) as f:
+                self.templates = yaml.safe_load(f)
+        else:
+            self.templates = index_path
 
         if not isinstance(self.templates, list):
             raise ValueError("Invalid templates.yaml, must be a list")
```
frazane: `self.templates` does not necessarily come from a YAML file (also below)

src/anemoi/inference/grib/templates/file.py
```diff
@@ -8,21 +8,36 @@
...
 class FileTemplates(TemplateProvider):
     """Template provider using a single GRIB file."""
 
-    def __init__(self, manager: Any, path: str) -> None:
+    def __init__(
+        self,
+        manager: Any,
+        *,
+        path: str,
+        mode: Literal["auto", "first", "last"] = "first",
```
frazane: Why is the default not "auto"?

src/anemoi/inference/grib/templates/file.py
```diff
@@ -31,23 +46,40 @@ def __init__(self, manager: Any, path: str) -> None:
...
 
-    def template(self, grib: str, lookup: dict[str, Any]) -> ekd.Field:
-        """Retrieve the template from the GRIB file.
+    def __repr__(self):
+        info = f"{self.__class__.__name__}({self.path.name},mode={self.mode}{{variables}})"
+        return info.format(variables=f",variables={self.variables}" if self.variables else "")
 
-        Parameters
-        ----------
-        grib : str
```
frazane: What do we gain by making this a cached property as opposed to just assigning an attribute inside the `__init__` (we would also know earlier if e.g. something goes wrong reading the data)?

EDIT: it's pointing to the wrong lines, the comment refers to the `_data` property.

src/anemoi/inference/grib/templates/manager.py
frazane: > ... to manage GRIB templates providers.

src/anemoi/inference/grib/templates/file.py
```diff
@@ -8,46 +8,79 @@
...
 import logging
+from functools import cached_property
+from pathlib import Path
 from typing import Any
+from typing import Literal
 
 import earthkit.data as ekd
 
+from anemoi.inference.decorators import main_argument
+from anemoi.inference.inputs.ekd import find_variable
```
HCookie: Arguably this should be moved into a utils

src/anemoi/inference/grib/templates/input.py
```diff
@@ -0,0 +1,68 @@
...
+        fallback : dict[str, str]
+            A mapping of output to input variable names to use as templates from the input,
+            used as fallback when the output variable is not present in the input state (e.g., for diagnostic variables).
+        """
+        super().__init__(manager)
+
+        if fallback:
+            self.fallback = fallback
+        else:
+            self.fallback = kwargs
```
HCookie: This will ignore kwargs if the fallback is set, I would suggest a `.update(kwargs)`?

src/anemoi/inference/grib/templates/manager.py
```diff
@@ -47,9 +50,33 @@ def __init__(self, owner: Any, templates: list[str] | str | None = None) -> None
...
+        to_log = {}
+
+        for provider, typed_list in self._history.items():
+            variables = [
+                variable for variable in typed_list if not getattr(variable, "_template_manager_logged", False)
+            ]
+            if len(variables) > 0:
+                to_log[provider] = set(variable.param for variable in variables)
+                for variable in variables:
+                    variable._template_manager_logged = True
```
HCookie: I'm not sure I like this setting of a private attribute on this object, Could you not use a set of param to unique identify and filter?

gmertes: Yep good point, done.

Issue #402, opened 2025-12-12
Support for multiple dataset checkpoints


gmertes:
We need to support checkpoints that have the new multi dataset format (ecmwf/anemoi-core#594).

The goal is to do it backwards compatible, and support both legacy single-dataset and new multi-dataset style checkpoints. The plan is to do it in two steps. Step 1 will be merged together with the PR in core to maintain feature parity, step 2 is a longer term task.

### Step 1
Provide support for multi-dataset checkpoints, but only those trained with a single dataset in and out. This is equivalent to the current style of checkpoints, except that dataset related metadata is now contained under a new default dictionary key `data`.

Supporting this is straightforward: we only have to map the old metadata to the new `data` entry, the contents of the metadata remain the same. For this we create a wrapper around the current Metadata class that redirects all existing attributes to the new `data` entry. For the rest of the code this is transparant and fully backwards compatible.

This allows us to merge the PR in core and not lose functionality in inference while we work on step 2. 

### Step 1.5
Together with the multi dataset checkpoints, we are also updating the checkpoint metadata with better inference metadata. Instead of 'reverse engineering' things like the name_to_index or the list of prognostic variables from the data indices, we can now just get them directly from the metadata.

We can start using these new entries in the wrapper described in step 1, updating the old attributes where applicable.

### Step 2 
Support the "real" multi-datasets case of N datasets in and out (e.g.: cerra + era in and out). This requires a refactor of all the logic related to input and output handling, as well as the rollout loop. We will need N inputs and outputs. For every tensor we currently create of each variable class (prognostic, diagnostic or forcing), we will now have to create them N times. 

We plan the creation of a new object, provisionally called the `TensorHandler`. The `TensorHandler` contains all the logic related to input, output and creation of tensors for a single dataset. We create N `TensorHandler`, and each one only has a view into the metadata for its corresponding dataset. This is where we use the Metadata wrapper from Step 1. The `TensorHandler` knows nothing about multi-datasets, it only knows about a single dataset. 

In the runner, we then call all N `TensorHandler` objects to prepare the input tensor of each dataset and do the rollout. 

Extra things to consider:

- The `State` will also have to be reworked. We have two options: we add an extra dictionary level to the current State object, one entry for each dataset, or we create a new multi dataset state object, that contains N State objects.
- The config will also need an extra level to specify separate inputs and outputs per dataset. How we do this is TBD, the simplest way is to add yet another level of dictionary.
[...]

PR #403, opened 2025-12-15
feat(multi-dataset): Support new checkpoints with single dataset name `data`

gmertes:
## Description
Implements step 1 (checkpoint support) and part of step 1.5 (new metadata) of #402 

Add support for the new multi-dataset checkpoints, but **only** the single-dataset case where the dataset name defaults to "data". Real multi dataset support will follow later in step 2.

This PR should be merged before ecmwf/anemoi-core#594 , but does not need to be merged exactly at the same time. This PR is forwards and backwards compatible.

I started using some of the new metadata but not all, to keep things simple. Improvements to the metadata will also be deferred to the next round.

### DIFF
--- a/src/anemoi/inference/metadata.py
+++ b/src/anemoi/inference/metadata.py
@@ -1171,7 +1171,7 @@ class MultiDatasetMetadata(Metadata):
         self.name = name
 
     def __repr__(self) -> str:
-        return f"<MultiDatasetMetadata name={self.name}>"
+        return f"MultiDatasetMetadata(name='{self.name}')"
 
     @property
     def _indices(self) -> DotDict:


src/anemoi/inference/metadata.py
```diff
@@ -1149,6 +1155,121 @@ def merge(main: dict[str, Any], patch: dict[str, Any]) -> None:
         merge(self._metadata, patch)
 
 
+class SingleDatasetMetadata(Metadata):
+    """Legacy single-dataset metadata."""
+
+    pass
+
+
+class MultiDatasetMetadata(Metadata):
```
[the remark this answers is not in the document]
gmertes: In part, yes. This class does two things:
1) Redirect all top level metadata entries to the new `data` key, so that all properties in the base class, derived from the original metadata, will still work.
2) Start using the new metadata under `metadata_inference`, all attributes below [this](https://github.com/ecmwf/anemoi-inference/pull/403/changes/BASE..2ef90bad30a24abdc87e94740d3f3cc2b676f0c6#diff-38579504182282fb338328fa4ddb04f1f43a53dc6a83a9ac1bce2e95737620e0R1196) line are "new": they are overrides for existing attributes.

The comment about "new metadata" refers to point 2. The base Metadata class still has many attributes that are not implemented in `MultiDatasetMetadata` , meaning they come from the base class (the original metadata). Eventually, we should aim to have all inference metadata derived from the new metadata, and all attributes implemented under `MultiDatasetMetadata`. 

Once we have that, the `MultiDatasetMetadata` could become the new base Metadata class, and we could deprecate the old one (or rename it to LegacyMetadata or something to keep for backwards compatibility). 

The reason I did not do that yet is because some code paths (and I think FIAB) imports `Metadata` to do type checking. So that's why for now the "old" metadata is still the base class, and we put the "new" one on top.

src/anemoi/inference/metadata.py
```diff
@@ -1149,6 +1155,121 @@ def merge(main: dict[str, Any], patch: dict[str, Any]) -> None:
...
+    """Map metadata for a multi-dataset checkpoint to a specific dataset name."""
+
+    multi_dataset = True
+
+    def __init__(self, metadata: dict[str, Any], supporting_arrays: dict[str, dict[str, FloatArray]] = {}, name="data"):
+        super().__init__(metadata, supporting_arrays.get(name, {}))
+        self.name = name
+
+    def __repr__(self) -> str:
+        return f"<MultiDatasetMetadata name={self.name}>"
```
HCookie: That is a little abnormal for a repr, is this standard across the Metadata class?

src/anemoi/inference/metadata.py
```diff
@@ -1149,6 +1155,121 @@ def merge(main: dict[str, Any], patch: dict[str, Any]) -> None:
...
+                raise ValueError(f"Variable {name} has no category")
+
+            result[name] = sorted(result[name])
+
+        self._variables_categories = frozendict(result)
+        return self._variables_categories
+
+
+class MetadataFactory:
+    def __new__(self, metadata: dict[str, Any], supporting_arrays: dict[str, Any] = {}, name="data") -> Metadata:
```
HCookie:
```suggestion
    def __new__(cls, metadata: dict[str, Any], supporting_arrays: dict[str, Any] = {}, name="data") -> Metadata:
```
I would also use kwargs for this, or at least for name

Issue #398, opened 2025-12-10
Time interpolator runner


mc4117:
### What happened?

The time interpolator runner currently has a couple of bugs in it. 

1. Forcings are expected from the forecasts but are not routinely saved in the forecasts and are not available in real time
2. Multistep_input is requested in several places but is not actually used in the time interpolator in anemoi-core
3. Runner does not currently support multi-output steps

### What are the steps to reproduce the bug?

-

### Version

-

### Platform (OS and architecture)

all

### Relevant log output

```shell

```

PR #412, opened 2026-01-06
fix: time interpolator for grib and forecast data

mc4117:
## Description
Steps to improve generalisability of time interpolator in inference and fix some bugs.

Allow inference with time interpolator to work from both grib files and mars (currently it only works from datasets).
This PR also fixes bugs relating to how forcings are handled which means that time-interpolator inference can be run in real time

Issue #398 


src/anemoi/inference/forcings.py
```diff
@@ -171,6 +171,52 @@ def rename(field: ekd.Field, _: str, metadata: dict[str, Any]) -> str:
...
+            The list of dates for which to load the forcings.
+        current_state : State
+            The current state of the model.
+
+        Returns
+        -------
+        FloatArray
+            The loaded forcings as a numpy array.
+        """
+        LOG.debug("Adding dynamic forcings %s", self.variables)
```
HCookie:
```suggestion
        LOG.debug("Adding computed forcings for interpolation %s", self.variables)
```

src/anemoi/inference/inputs/ekd.py
```diff
@@ -178,13 +180,25 @@ def _name(field: Any, _: Any, original_metadata: dict[str, Any]) -> str:
...
 
-        # for f in data:
-        #     LOG.info("Field %s %s", f.metadata("name"), f.metadata("valid_datetime"))
+        start_date = kwargs.pop("start_date", None)
 
-        data = data.sel(name=self.variables, valid_datetime=valid_datetime).order_by(
-            name=self.variables,
-            valid_datetime="ascending",
-        )
+        # Deselects analysis if the forecast exists to avoid duplicate values for same valid time
```
OpheliaMiralles: Does this assume that the path to analysis files and the path to the forecasts is the same? Couldn't it be solved by saving the forecasts in a different folder and running the interpolator on data from this folder? This is how we did it at MeteoSwiss and seemed to work

OpheliaMiralles (commented):
I would just like to understand the context in which start_date is added, and if the specific case where analysis and forecasts are in the same input search path needs to be handled in the code.

PR #463, opened 2026-03-17
feat: Full multi-dataset support

gmertes:
# Full Multi-Dataset Support

Apologies for the wall of text, but this is a big one and I suggest you read everything before reviewing. 

Implements **Step 2** of #402: full support for checkpoints trained on N datasets, where each dataset can have its own variables, inputs, and outputs.

While the diff touches a lot of files, many of the changes are small and mechanical, the most common being the replacement of `checkpoint` with a per-dataset `metadata` object in all inputs, outputs, pre-processors, and post-processors. 

The biggest changes are in: `checkpoint.py`, `metadata.py`, `runner.py`, `tensors.py`.

The main architectural changes are described below.

## Checkpoint / Metadata Separation

The `Checkpoint` and `Metadata` classes now have clearly separated responsibilities:

- **`Checkpoint`** represents the file on disk and exposes only attributes that are **shared across all datasets** (e.g., `timestep`, `multi_step_input`, `multi_step_output`, `precision`, `lagged`). It no longer contains dataset-specific metadata redirects, these have all been removed.
- **`Metadata`** now only carries **dataset-specific** metadata. 
- `Checkpoint.multi_dataset_metadata` returns a `dict[str, Metadata]`, one entry per dataset, which is the main entry point for iterating over datasets and metadata.

Any object that is initialised per dataset (inputs, outputs, processors) receives its own dataset-scoped `Metadata` instance rather than the checkpoint.

## Runner Consolidation

All functionality from `DefaultRunner` has moved into the base `Runner` class. The `Runner` now takes a `RunConfiguration` as a required positional argument. Subclasses that don't work from a config (like `SimpleRunner`) create a hidden config. In the case of `SimpleRunner`, a config with empty inputs and outputs.

The reason for this is that there were methods in `Runner` that implicitly relied on having a config (like creation of forcings inputs), but `Runner` did not have the config. They would only work from the config-driven `DefaultRunner`. This made it hard to move code from `Runner` to `TensorHandler`, because the `TensorHandler` is initialised in `Runner`, but the `TensorHandler` needed access to the config in order to inialise the forcing inputs. If that sounds confusing, that's because it was 😅. This change simplifies the class-hiarchy and makes separation of concerns clear.

## Inputs, Outputs, and Processors as Runner Attributes

Inputs, outputs, pre-processors, and post-processors are now created in the `Runner` constructor, and we create one of each per dataset (even if they are empty). They are now stored as instance attributes, in dictionaries keyed by dataset name:

```python
self.pre_processors: dict[str, list[Processor]]
self.post_processors: dict[str, list[Processor]]
self.prognostics_inputs: dict[str, Input]
self.constant_forcings_inputs: dict[str, Input]
self.dynamic_forcings_inputs: dict[str, Input]
self.boundary_forcings_inputs: dict[str, Input]
self.outputs: dict[str, Output]
```

Previously, these objects were created in different places (and often multiple times), making it hard to trace their origins. Having them as constructor-initialised attributes makes the data flow explicit.

## Tensors as Dictionaries

In the `Runner`, everything that used to be a single tensor is now a dictionary of tensors keyed by dataset name. The same goes for the State. This applies to:

- Input tensors passed to `predict_step`
- Output tensors returned from `predict_step`
- All tensors internal to the rollout loop in `forecast()`
- State objects passed through `run()` and `execute()`

## TensorHandler

A new `TensorHandler` class (`tensors.py`) encapsulates all tensor preparation, forcings loading, and rollout tensor update logic **for a single dataset**. This class does not do anything new: all functions here were copied from the `Runner`. Just like inputs and outputs, the Runner creates one `TensorHandler` per dataset and orchestrates them during the forecast loop. 

Note that in contrast to the `Runner`, the `TensorHandler` does not deal with dictionaries. Internally, everything is a single tensor (like it used to be in the `Runner`). Each handler only knows about a single dataset. The Runner handles the multi-dataset coordination.

## Backwards compatibility
Everything is backwards compatible with legacy checkpoints trained before multi-datasets. This is largely handled by the `MetadataFactory`. When it detects a legacy model, it returns the metadata with the hardcoded dataset name `data`. This is functionally the same as a multi-dataset checkpoint with only one dataset. 

When `predict_step` detects a legacy checkpoint, it unwraps the input tensor dictionary to a single tensor before calling the model and re-wraps the output. 

The config is also fully backwards compatible and does not change for single-dataset checkpoints.

## Config changes

The relevant config entries can now take dataset-specific entries, by placing them in a dictionary keyed by dataset name. This allows users to specify different inputs, outputs, and processors per dataset. Entries that now take multi-dataset configuration are: `input`, `output`, `pre_processors`, `post_processors`.

```yaml
output:
  era5:
    grib: output-era5.grib
  cerra:
    netcdf: output-cerra.nc
```

In the case of single-dataet checkpoints, you don't have to include the extra dictionary with dataset name. You can just use the old format. **If no per-dataset configuration is provided, the config is reused for all datasets.** So you can still do this:

```yaml
input:
  grib: input.grib
```

***WARNING:*** When you use the above with a checkpoint that has >1 datasets, **a separate GRIB input will be created for each dataset**, and each one will read `input.grib` separately. This is **NOT** recommended (see "out of scope section" below), I'm just highlighting it here to illustrate the behavior.

A new `@format_dataset_name` decorator substitutes `{dataset}` placeholders in string arguments with the actual dataset name from `metadata.dataset_name`.

For example, the following will automatically create two netcdf outputs with `output-era5.nc` and `output-cerra.nc`:

```yaml
output:
  netcdf: output-{dataset}.nc
```

The above is equivalent to:

```yaml
output:
  era5:
    netcdf: output-era5.nc
  cerra:
    netcdf: output-cerra.nc
```

You MUST create separate output files for each dataset. Writing to the same file is not supported (see out of scope below).

## How to test this
In addition to testing with legacy single-dataset checkpoints like AIFS 1.1, I have been testing this by running the 3in-2out era-cerra checkpoint from the anemoi-training integration test. For inputs I have tested: mars, dataset and grib. For outputs I tested raw, netcdf, grib.

For exampple, the following config works:

```yaml
checkpoint: era-cerra-3in-2out-from-integration-test.ckpt
input: training
output:
  era5:
    grib: output-era5.grib
  cerra:
    netcdf: output-cerra.nc
```

With a real checkpoint you could test by running from `input: test`, which will run each dataset for their corresponding test dataset. Or you could run from grib/netcdf inputs if you have them, but note that in this case you have to do something like:

```yaml
# assuming you have input-[datasetA].grib and input-[datasetB].grib
input:
  grib: input-{dataset}.grib

# or more exlicitly something like
input:
  datasetA:
    grib: some-input.grib
  datasetB:
    netcdf: another-input.nc
```

## Other Notable Changes

- **`RunnerClasses` configuration**: A new Pydantic model (`RunnerClasses`) allows subclasses to swap implementations of `TensorHandler`, `Checkpoint`, and `Metadata`.
- **Interpolator updated for multi-dataset**: with a custom `TimeInterpolatorTensorHandler` and `TimeInterpolatorMetadata`. This is a good example of how child runners can now define their own tensor handlers and metadata with `RunnerClasses`, instead of monkey patching.
- **Mock model updated for multi-dataset**: `SimpleMockModel` used by the integration test now behaves like a multi-dataset model, while still supporting single-tensor legacy mode.
- **Checkpoint slimmed down**: dataset-specific attribute redirects that are now handled by the `Metadata` have been removed.
- **`ai-models` integration removed**. This was a legacy integration that is no longer supported.
- Removed redundant docstrings of internal methods in cases where the function signature was entirely self-documented.

## Work in Progress

The PR is opened as a draft as I'm still working on the following, but the main functionality can be reviewed and tested.

- [x] Document the config changes
- [x] Update integration test to work with >1 dataset models
- [x] Add >1 dataset model to the integration test
- [x] Multi-dataset support for `patch_metadata`
- [x] Multi-dataset support for `trace`
- [x] Multi-dataset support for `retrieve`. This one is ecmwf/prepml specific, so we can also update it in a separate PR.

## Out of scope for this PR

- **Multi-dataset support for `external_graph` runner.** Let's do it in another PR as this one is already huge 💀
- Same with the coupled runner, fix in another PR. 
- **We need to revisit the contents of the `Context` ABC.** This PR greatly reduces reliance on the context to pass information around, and I think the ABC is not useful anymore. Leave as-is for now, revisit later. 
- **Writing outputs to the same file is not supported.** If there are multiple datasets, we assume that the output files are also different. 
- **Having separate input files per dataset is recommended.** Reading input from the same file/dataset should work (I've not tested this), but is not optimised because each input will open and read the file separately. For now, we assume you have a separate input file/source per dataset. 
- **GRIB output**. GRIB output only works out-of-the-box for global models with grids that have built-in templates (like O96, N320). For other grids like CERRA, and for LAM/SGM, the user has to provider their own GRIB templates using the template manager. I've not tested this with a >1 dataset checkpoint. I suggest you use NetCDF (that's what I tested and can confirm working for CERRA models). Good luck and godspeed if you're a GRIB user.

----
📚 Documentation preview 📚: https://anemoi-inference--463.org.readthedocs.build/en/463/

### DIFF
--- a/src/anemoi/inference/decorators.py
+++ b/src/anemoi/inference/decorators.py
@@ -181,6 +181,10 @@ class format_dataset_name:
         if not any("metadata" in inspect.signature(klass.__init__).parameters for klass in cls.mro()):
             raise TypeError(f"`{self.__class__.__name__}` can only be used to decorate classes that take `metadata`")
 
+        class DefaultFormat(dict):
+            def __missing__(self, key):
+                return f"{{{key}}}"  # if the key is not found, return the placeholder unchanged
+
         class WrappedClass(cls):
             def __init__(wrapped_cls, context: Context, metadata: Metadata, *args: Any, **kwargs: Any) -> Any:
                 assert self.arg in kwargs, f"{self.arg} not found in decorated class arguments: {kwargs}"
@@ -189,7 +193,7 @@ class format_dataset_name:
                 ), f"{self.arg} must be a string to use `{self.__class__.__name__}` decorator"
 
                 name = metadata.dataset_name
-                kwargs[self.arg] = kwargs[self.arg].format(dataset=name, dataset_name=name)
+                kwargs[self.arg] = kwargs[self.arg].format_map(DefaultFormat(dataset=name, dataset_name=name))
                 super().__init__(context, metadata, *args, **kwargs)
 
         return type(cls.__name__, (WrappedClass,), {})
--- a/tests/unit/test_decorators.py
+++ b/tests/unit/test_decorators.py
@@ -68,6 +68,9 @@ def test_format_dataset_name():
     cls = _Cls("context", metadata, path="output-{dataset}.grib")
     assert cls.path == "output-era5.grib"
 
+    cls = _Cls("context", metadata, path="output-{dataset}-{levtype}.grib")
+    assert cls.path == "output-era5-{levtype}.grib"
+
     cls = _Cls("context", metadata, path="output-{dataset_name}.grib")
     assert cls.path == "output-era5.grib"
 


src/anemoi/inference/inputs/empty.py
```diff
@@ -43,8 +44,7 @@ def __init__(self, context: Context, **kwargs: Any) -> None:
         **kwargs : object
             Additional keyword arguments.
         """
-        super().__init__(context, **kwargs)
-        assert self.variables in (None, []), "EmptyInput should not have variables"
+        super().__init__(context, metadata, **kwargs)
```
mc4117: just wondering, why has the assert disappeared?

tests/integration/test_integration.py
```diff
@@ -166,6 +166,7 @@ def load_array(task):
 
 def test_integration(test_setup: Setup, tmp_path: Path) -> None:
     """Run the integration test suite."""
+    # TODO: full multi-dataset support: integration test checkpoints can currently only have 1 dataset
```
mc4117: Do the system level tests also need to be updated?

src/anemoi/inference/runners/interpolator.py
```diff
@@ -158,123 +167,80 @@ def execute(self) -> None:
...
-                if self.trace:
-                    self.trace.from_source(name, source, "initial dynamic forcings")
+            # Run interpolation for this window
+            for state_idx, states in enumerate(
+                self.run(input_states=input_states, lead_time=self.interpolation_window)
+            ):
+                for dataset, state in states.items():
+                    # In the first window, we want to write the initial state (t=0)
+                    # In other windows, we want to skip the initial state (t=0)
+                    # because it is written as the last state of the previous window
```
mc4117: I actually was thinking about this and maybe we need to make this a flag. Because we don't always have to predict the last step

src/anemoi/inference/runners/interpolator.py
```diff
@@ -98,53 +149,11 @@ def patch_data_request(self, request: dict) -> dict:
         )
         return req
 
-    def _patch_checkpoint_lagged_property(self):
```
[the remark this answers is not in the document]
gmertes: Ah good point, not with a unit test but it's implicitly tested by the integration test. It will fail if it was looking backwards, because the input grib only has data in the future from the specified start date. I confirmed that using `Metadata` instead of `TimeInterpolatorMetadata` makes it fail like this. Then there is an extra sanity check to see that the output time step and valid time only increases in time.

docs/inference/configs/multi-dataset.rst
```diff
@@ -0,0 +1,142 @@
...
+The same pattern applies to inputs:
+
+.. code:: yaml
+
+   input:
+     era5:
+       mars:
+         class: ea
+     cerra:
+       grib: cerra-input.grib
```
HCookie: I see that this is inline with the examples in training configs, so maybe this is more of a general comment, but I think the definition of era5 and cerra break down in the inference context. It would make more sense to have global and local? It's a semantic thing I know, but could we have a remapping?

src/anemoi/inference/checks.py
```diff
@@ -12,33 +12,33 @@
 import logging
 import sys
 from collections import defaultdict
-from typing import Any
 
 from anemoi.utils.humanize import plural
+from earthkit.data.indexing.fieldlist import FieldArray
```
HCookie:
```suggestion
from earthkit.data import FieldList
```

src/anemoi/inference/tensors.py
```diff
@@ -0,0 +1,611 @@
...
+        LOG.info("  Dynamic forcings:")
+        for f in initial_dynamic_forcings_inputs:
+            LOG.info(f"    {f}")
+        LOG.info("Initial forcings dates:")
+        LOG.info(f"  {', '.join([date.isoformat() for date in dates])}")
+        LOG.info("-" * 80)
+
+        for source in initial_constant_forcings_inputs:
+            arrays = source.load_forcings_array(dates, input_state)
+            for name, forcing in zip(source.variables, arrays):
```
HCookie: Not sure if this is new or not, but `variables` is not a guaranteed attr on forcings

src/anemoi/inference/decorators.py
```diff
@@ -76,6 +90,7 @@ class ensure_path:
     If `is_dir` is True, the path is treated as a directory, if not for files, the parent directory is treated as a directory.
     If `must_exist` is True, the directory must exist.
     If `create` is True, the directory will be created if it doesn't exist.
+    If 'unique' is True, the same path cannot be reused between multiple decorated classes.
```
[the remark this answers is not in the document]
gmertes: Yeah so this is the mechanism that requires all output paths to be unique, to avoid clashing outputs. This decorator is on all outputs, and I default unique=True

frazane:
Amazing work @gmertes! We (@clairemerker) ran a few tests on our end and we are able to run inference with this branch. However, we are having some issues with the GRIB output, related to (who would have thought) variables metadata. With some metadata patching, we still managed to write some GRIB output, but something might not be encoded correctly. We are still investigating but in general this looks good already, considered you explicitly mention GRIB is not expected to work out of the box for our use case ;-)

gmertes:
Nice, thanks for testing! How do you want to proceed for your use-case? If the variables metadata issue is on the inference side I'm happy to have a look at fixing it in this PR. But if you say it can be fixed with patching, it could come from the training/dataset side. 

In theory each dataset's grib output is independent and identical to before, so as long as the variables metadata is the same as before (and correct) it should™ work (but that's not accounting for more complicated setups like using the cutout input and extract_from_state processor).

Feel free to send me some logs on Slack, happy to have a look.

HCookie (dismissed):
Amazing works, thanks.

src/anemoi/inference/decorators.py
```diff
@@ -130,5 +157,39 @@ def __init__(context, dir=None, ...):
...
+
+        class WrappedClass(cls):
+            def __init__(wrapped_cls, context: Context, metadata: Metadata, *args: Any, **kwargs: Any) -> Any:
+                assert self.arg in kwargs, f"{self.arg} not found in decorated class arguments: {kwargs}"
+                assert isinstance(
+                    kwargs[self.arg], str
+                ), f"{self.arg} must be a string to use `{self.__class__.__name__}` decorator"
+
+                name = metadata.dataset_name
+                kwargs[self.arg] = kwargs[self.arg].format(dataset=name, dataset_name=name)
```
JPXKQX: I think this will fail if the keyword argument is a path with more keys than the formatted string. For example, if you pass a path like `"output-{dataset}-{extra}.grib"`.

JPXKQX: I am happy for this to be fixed later, I just want to raise it

gmertes: Ohh thanks for catching this! Fixed it in the latest commit.

PR #466, opened 2026-03-19
fix: use inference metadata to get explicit times

VeraChristina:
## Description
This [PR in training](https://github.com/ecmwf/anemoi-core/pull/896) will remove `explicit_times` from the training config for the interpolator. Use the inference metadata instead to get the explicit times for inference. 

Note that all multi-out interpolator checkpoints are multiple-dataset checkpoints so the inference metadata is available for these checkpoints.

## Additional notes
For `multi_step_input` and `multi_step_output` we already use the inference metadata. So the changes to the training config for the forecaster don't require additional changes in inference.

BEGIN_COMMIT_OVERRIDE
fix(metadata): use inference metadata to get explicit times (#466)
END_COMMIT_OVERRIDE

### DIFF
--- a/.github/workflows/inactivity-bot.yml
+++ b/.github/workflows/inactivity-bot.yml
@@ -5,13 +5,19 @@ on:
     - cron: "0 23 * * *"  # every day at 23pm on default(main) branch
   workflow_dispatch: # Allows manual trigger
 
+permissions:
+  actions: write
+  contents: write # only for delete-branch option
+  issues: write
+  pull-requests: write
+
 jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/stale@v9
         with:
-          repo-token: $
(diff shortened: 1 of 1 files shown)


src/anemoi/inference/metadata.py
```diff
@@ -1253,6 +1253,16 @@ def multi_step_output(self) -> int:
...
 
+    @cached_property
+    def input_explicit_times(self) -> Any:
+        """Explicit times of the input steps used for the interpolator."""
+        return self._inference.timesteps.input_relative_date_indices
+
+    @cached_property
+    def target_explicit_times(self) -> Any:
+        """Explicit times of the target steps used for the interpolator."""
+        return self._inference.timesteps.output_relative_date_indices
```
HCookie: Where are these to be used? or is that coming later?

PR #491, opened 2026-04-17
feat: allow using the filter's forward method in forward_transform_filter

bdvllrs:
## Description

Currently the `forward_transform_filter` uses the backward method of a filter and applies a `.reverse()`. This adds the argument `use_forward` to allow using the forward method of the filter instead. This can be useful for filters like `regrid`. (see ecmwf/anemoi-transform#275).

Some forward filters (like regrid) would be useful as post-processors but are not currently accessible as they don't implement a backward method.

Related to ecmwf/anemoi-transform#275

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-inference--491.org.readthedocs.build/en/491/

### DIFF
--- a/docs/inference/configs/processors.rst
+++ b/docs/inference/configs/processors.rst
@@ -22,8 +22,7 @@ Replaces NaNs with the mean.
 forward_transform_filter
 ========================
 
-Applies a filter from `anemoi-transform
-<https://anemoi.readthedocs.io/projects/transform/en/latest/_api/transform.filters.html>`_.
+Applies a filter from :ref:`anemoi-transform <anemoi-transform:list-of-filters>`.
 
 Either the name of the filter can be provided as a string or a dictionary with the filter name as the key and the filter arguments as the value. For example:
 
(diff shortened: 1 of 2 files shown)


docs/inference/configs/processors.rst
```diff
@@ -122,6 +122,23 @@ backward_transform_filter
...
+
+For example to use regrid as a post-processor:
+
+.. code:: yaml
+
+   post_processors:
+     - forward_transform_filter:
+         use_forward: true
+         filter: regrid
+         # ...
```
gmertes: Thanks! This looks good, I only suggest we document the new config format introduced recently, which is more consistent with the rest of the config style:
```suggestion
     - forward_transform_filter:
         regrid:
           use_forward: true
           # other options for the filter
```

docs/inference/configs/processors.rst
```diff
@@ -122,6 +122,23 @@ backward_transform_filter
 Applies a backward transform filter from `anemoi-transform
 <https://anemoi.readthedocs.io/projects/transform/en/latest/_api/transform.filters.html>`_.
 
+forward_transform_filter
+=========================
+
+Applies a backward transform on the reversed filter from `anemoi-transform
+<https://anemoi.readthedocs.io/projects/transform/en/latest/_api/transform.filters.html>`_
```
gmertes: The link is broken, I guess it is now:
`https://anemoi.readthedocs.io/projects/transform/en/latest/filters/list-of-filters.html`

Or better would be to link to the specific sphinx label, I think something like this should work:
``` :ref:`anemoi-transform <anemoi-transform:list-of-filters>` ```

Do you want to update it in the rest of the page too?

bdvllrs: Yes, I can do that

bdvllrs:
With 18a44d5 I removed some duplicated code.

gmertes (approved):
Nice 👌

PR #496, opened 2026-04-23
feat(patch): Support for multi-dataset checkpoint patching

### LINKED ISSUE #486: Update `patch` command for multi-dataset checkpoints
The patch command edits the raw metadata dictionary, in particular it replaces the top level `dataset` entry. It doesn't account for the new multi-dataset format where datasets are nested under their own entry.

Update `patch` to loop over the datasets and replace the correct `dataset` sub-entry.

gmertes:
## Description
Title. Requires ecmwf/anemoi-utils#288 to be merged and released first.

Got rid of the convoluted patch mixin that didn't fit anymore in the new multi-dataset/multi-metadata design. It's now a single method on the Checkpoint class that deals with both single and multi-dataset cases. And it's no longer called patch_metadata because that is already overloaded with the one from the config. 

Also updated the return type hints on the metadata factories with the concrete classes rather than the base class.


HCookie (approved):
LGTM

PR #498, opened 2026-04-28
perf(forcings): Skip loading of initial forcings already in input state

gmertes:
## Description
We've had a long standing issue of loading the initial forcings twice: once when we create the input state from the input objects, and then again in the subroutine `add_initial_forcings_to_input_state()` where we ask the forcing providers to load their forcings into the input state. 

This subroutine exists because of coupling: the coupled runner may need to inject different initial forcings. But there is no reason to be loading initial forcings that were already loaded by the input(s), even in a coupled scenario (at least currently there is no use-case to do so). So we can skip these entirely.

The performance improvement is marginal, but the nice side effect is that it simplifies the temporal downscaler: the custom tensorhandler is no longer needed because that only existed to deal with the initial constant forcings. Constant forcings will now automatically be loaded only once in the first input state and carried forward throughout the run. 

### DIFF
--- a/src/anemoi/inference/forcings.py
+++ b/src/anemoi/inference/forcings.py
@@ -234,36 +234,6 @@ class CoupledForcings(Forcings):
         )
 
 
-class ConstantDateForcings(CoupledForcings):
-    """Retrieve forcings from the first date in the input. Used, for example, in the temporal downscaler where forcings are only available at the first time step of the input forecast."""
-
-    def load_forcings_array(self, dates: list[Date], current_state: State) -> FloatArray:
-        """Load the forcings for the given dates.
-
-        Parameters
-        ----------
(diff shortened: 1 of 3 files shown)


mc4117:
Looks good! Would it be possible to test this by applying the temporal downscaler to a grib file without forcings in the forecast? (it's possible this is what the system level tests do)

VeraChristina:
Nice PR!

> Looks good! Would it be possible to test this by applying the temporal downscaler to a grib file without forcings in the forecast? (it's possible this is what the system level tests do)

The grib file used for the temp downscaler integration test has lsm/sdor/slor/z only at step=0.

mc4117 (approved):
Thanks for the great work! LGTM

HCookie (approved):

PR #501, opened 2026-05-07
feat(outputs): parallel writing output

cathalobrien:
## Description
This PR adds parallel output support to anemoi inference. The process attached to GPU 0 spawns multiple writer processes. It splits the output along the field dimension and each writer writes a unique output

Very similar to the `tee` output, the `parallel` output takes a number of writer processes and wraps another output class
```
output:
  parallel:
    num_writers: 4
    output:
      grib:
        path: $HPCPERM/aifs/tests/par-inf/out/writers4/out.grib
checkpoint: $SCRATCH/aifs/n320/checkpoint/cfc98c50-0114-4af1-b304-bfc347c9e931/inference-last.ckpt
input:
  grib:  $PERM/aifs/32km/aifs-input-inference-last.grib 
lead_time: 240
```

the writers id will be added to the end of the filename. The above example will create the following outputs:
```
ls out/writers4/
  out_w0.grib  out_w1.grib  out_w2.grib  out_w3.grib
```

I added documentation [here](https://anemoi--501.org.readthedocs.build/projects/inference/en/501/inference/parallel_output.html)

----
📚 Documentation preview 📚: https://anemoi-inference--501.org.readthedocs.build/en/501/


frazane:
Could asynchronous writing also be in scope for this PR? As in allowing the autoregressive rollout to continue while the outputs of the previous steps are still being written.

cathalobrien:
> Could asynchronous writing also be in scope for this PR? As in allowing the autoregressive rollout to continue while the outputs of the previous steps are still being written.

@frazane yes, if you use 1 writer process then you will write a single output file asynchronously while the next forecasts are being computed 

I will update the docs to make this option a bit clearer

frazane:
Okay, at first I thought this PR parallelized writing a single state, e.g. different variables in parallel. It is in fact writing separate steps together, is that correct? 

If users already write different output steps as separate files, is appending `_{worker_id}` still done?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        # __init__ time the context may not yet have lead_time / time_step set.
+        # Forking before those are populated would give the children stale copies.
+        self._writers_running = False
+
+    def open(self, state: State) -> None:
+        """Spawn the writer processes during open() instead of __init__() to ensure they have access to the full context."""
+        if not self._writers_running:
+            self._spawn_writers(self.context, self.output_config)
+            self._writers_running = True
+        # self.write_state(state) # Do I need to write the initial state here?
```
gmertes: nope, comment can be removed

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        - output_w1.grib
+
+    """
+
+    def __init__(
+        self,
+        context: Context,
+        metadata: Metadata,
+        *,
+        output: Output | Any,
```
gmertes: A small QoL we could do is if `output` is not defined, we use the `kwargs` for the output, so you could simply write:

```yaml
output:
  parallel:
    num_writers: 4
    grib: out.grib
```
Although maybe we want the explicitness of `output`, @HCookie what do you think?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        if self.output_frequency is not None:
+            if (step % self.output_frequency).total_seconds() != 0:
+                return
+
+        if self.num_writers == 0:
+            # write in the main process without parallelism
+            return super().write_state(state)
+
+        # Parallel case — split state into chunks and send to writers via queues
+        state_chunks = _split_state(state, self.num_writers)
```
gmertes: Is this splitting deterministic, so that the same variables always end up in the same chunk (and a writer always gets the same variables) ? Idk if we need that, but it could be nice to have.

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def _spawn_writers(self, context: Context, output_config) -> None:
+        """Fork writer processes.
+
+        'self.num_writers' writer processes are spawned, each with its own queue for receiving states to write.
+        The writer processes run the '_writer_loop' method, which listens for incoming states on the queue and
+        calls 'write_step' on the wrapped output.
+        """
+        self._processes: list[mp.Process] = []
+        self._queues: list[mp.Queue] = []
```
gmertes, line 243: I think we should put a maximum size on the queues so we don't blow up the memory. There's no benefit to the model running too far ahead of the disk if at the end you still have to wait for the disk to catch up. 

I'd be conservative and choose a relatively low value, enough to absorb fluctuations in the disk but not too high so that it drastically increases memory usage. You could make it configurable, with a conservative default value?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+
+        output = create_output(context, output_config, self.metadata, suffix=f"_w{writer_id}")
+        has_written_initial_state = False
+
+        while True:
+            # Receive a message from the main process
+            try:
+                message = queue.get()
+            except mp.queues.Empty:
+                continue
```
gmertes: Is this still needed?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+                output.write_step(message)
+                LOG.debug("Writer %d done: %s", writer_id, message.get("date"))
+            except Exception as e:
+                LOG.error("Writer %d write error: %s\n%s", writer_id, e, traceback.format_exc())
+                break
+
+        LOG.info("Writer %d shutting down", writer_id)
+        exit(0)
+
+    def _terminate_all_writers(self) -> None:
```
gmertes: Does the process still terminate cleanly on an interrupt (SIGINT or SIGTERM) ?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+                    # each process needs to write initial state to do some setup
+                    output.write_initial_state(message)
+                    has_written_initial_state = True
+
+                LOG.debug("Writer %d writing: %s", writer_id, message.get("date"))
+                output.write_step(message)
+                LOG.debug("Writer %d done: %s", writer_id, message.get("date"))
+            except Exception as e:
+                LOG.error("Writer %d write error: %s\n%s", writer_id, e, traceback.format_exc())
+                break
```
gmertes, line 305: Does this logic mean that a worker could crash while the rest keeps running? If that's the case, should we kill everything if one of the workers throws an error?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+                message = self.post_process(message)
+
+                if not has_written_initial_state:
+                    # Some outputs (e.g. netcdf) must be explictly opened.
+                    if hasattr(output, "open"):
+                        LOG.debug("Writer %d openning %s", writer_id, output)
+                        output.open(message)
+                    # each process needs to write initial state to do some setup
+                    output.write_initial_state(message)
+                    has_written_initial_state = True
```
gmertes: Wait, what happened to the `write_initial_state` method on the ParallelOutput? Should we not drive this logic through that function? We could have the ParallelOutput dispatch a different message type to indicate it's the initial state (you could make the message a tuple with extra information attached to it).

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+
+                LOG.debug("Writer %d writing: %s", writer_id, message.get("date"))
+                output.write_step(message)
```
gmertes: What's the reason for calling write_step here instead of write_state ? Could we not reduce some code duplication by using write_state : the output_frequency and post_process could be handled by the child output?

gmertes (commented):
Very cool 🚀

src/anemoi/inference/outputs/parallel.py
cathalobrien: yeah its deterministic, 

Each writer gets a subset of the fields based on its rank, with the final writer getting the remainder.
```
fields_per_chunk = math.ceil(len(fields) / num_chunks)
start = i * fields_per_chunk
end = min(start + fields_per_chunk, len(fields))
```

src/anemoi/inference/outputs/parallel.py
cathalobrien: good idea

src/anemoi/inference/outputs/parallel.py
cathalobrien: no, good catch

src/anemoi/inference/outputs/parallel.py
cathalobrien: I tested with `pkill -2 anemoi*` and `pkill -9 anemoi*` and both exited cleanly

VeraChristina:
Thanks for adding this! I have done a small test run with a temporal downscaler. Variables are distributed across output files as expected. Comparing the grib files shows no missing fields in the writer files, but additional fields at step 0 (compared to the same inference run on main) for all non-accumulation variables. (Also, some metadata naming seems to differ but I would need to double check.)

I haven't looked further so it's possible that this can easily be disabled by setting write_initial_state to False or alike, but I thought it might be worth flagging.

To reproduce: run an temporal downscaler on the branch with the settings as in PR description, and on main with normal grib output. Happy to provide a checkpoint

src/anemoi/inference/outputs/parallel.py
cathalobrien: good idea, just a lack of understanding of the differences on my side

docs/inference/parallel_output.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Each writer process handles a subset of the output fields and writes
+them independently to its own file. This distributes both the I/O
+bandwidth and the encoding/serialisation cost across multiple CPU cores.
+
+***************
+ How It Works
+***************
+
+The ``parallel`` output wraps any other output type (e.g. ``grib``,
```
HCookie, line 24: Can I do parallel plots?

docs/inference/parallel_output.rst
```diff
@@ -0,0 +1,170 @@
...
+
+   -  -  ``num_writers``
+      -  Yes
+      -  Number of writer processes to spawn. Each writer handles an
+         equal share of the output fields.
+
+   -  -  ``output``
+      -  Yes
+      -  The inner output configuration. Can be any supported output
+         type (``grib``, ``zarr``, ``netcdf``, etc.).
```
HCookie: Can this not be autodoc generated?

docs/inference/parallel_output.rst
```diff
@@ -0,0 +1,170 @@
...
+       num_writers: 4
+       output:
+         grib:
+           path: /path/to/output.grib
+
+***************************
+ Choosing ``num_writers``
+***************************
+
+For small resolutions or infrequent output steps, paralel writing might be unnecessary.
```
HCookie:
```suggestion
For small resolutions or infrequent output steps, parallel writing might be unnecessary.
```

src/anemoi/inference/outputs/parallel.py
cathalobrien: Good point, 
I modified write_state() to raise a runtime error on encountering a dead process
```
if not self._processes[i].is_alive():
                raise RuntimeError(
                    f"Writer {i} is dead, inference will now fail. Check previous logs for errors in the writer process."
                )
```

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+    if isinstance(obj, torch.Tensor):
+        return obj.detach().cpu().numpy()
+    if isinstance(obj, dict):
+        return {k: _detach_tensors(v) for k, v in obj.items()}
+    if isinstance(obj, (list, tuple)):
+        return type(obj)(_detach_tensors(v) for v in obj)
+    return obj
+
+
+def _sanitize_state(state: State) -> State:
```
HCookie:
```suggestion
def _sanitise_state(state: State) -> State:
```

src/anemoi/inference/outputs/gribfile.py
```diff
@@ -421,3 +426,10 @@ def __init__(
             negative_step_mode=negative_step_mode,
             **kwargs,
         )
+
+    @staticmethod
+    def _add_suffix_to_output_path(path, suffix) -> Path:
+        """Add the suffix to the output path."""
+        extension = ".grib"
+        new_path = str(path).removesuffix(extension) + suffix + extension
+        return Path(new_path)
```
HCookie:
```suggestion
    @staticmethod
    def _add_suffix_to_output_path(path: Path, suffix: str) -> Path:
        """Add the suffix to the output path."""
        return path.with_name(path.name + suffix)
```

src/anemoi/inference/outputs/zarr.py
```diff
@@ -328,3 +331,13 @@ def close(self) -> None:
...
+
+    @staticmethod
+    def _add_suffix_to_output_store(store, suffix) -> "StoreLike":
+        """Add the suffix to the output store."""
+        extension = ".zarr"
+        new_store = str(store).removesuffix(extension) + suffix + extension
+        if TYPE_CHECKING:
+            return StoreLike(new_store)
+        else:
+            return new_store
```
HCookie:
```suggestion
        return new_store
```

HCookie (changes requested):
I haven't yet looked at the parallelisation aspect, I will do so soon.
I think we can limit the impact on the inputs a little further than this. 

A decorator on top of the inputs, that intercepts a magic suffix key and adds it to the path might be a good option. You could look at the `ensure_path` decorator for inspiration on this?

I'd see it taking an arg, which kwarg to add to, checking if it is a str or Path, and then pop-ing `parallel-suffix` from the kwargs.

```python
@supports_parallel("path")
class GribOutput:
	...
```

This would be an easy way to limit which outputs can be parallelised, and ensure no special functions are needed.
Happy to chat further on this, or provide a PR into this one?

src/anemoi/inference/outputs/parallel.py
HCookie: I do like the use of kwargs for this, would be nice and easy to use

cathalobrien: I have set that up now, I'll update the docs

src/anemoi/inference/outputs/gribfile.py
cathalobrien: good idea, I guess you mean 
```
suffix="_w0"
# path.suffix=.grib
# suffix+path.suffix=w0.grib
return path.with_suffix(suffix+path.suffix)
```
dual use of suffix matches it unnecessarily confusing

HCookie: pathlib won't let that as a suffix must start with `.`, but just changing the name works fine

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
+# (C) Copyright 2024 Anemoi contributors.
```
HCookie:
```suggestion
# (C) Copyright 2026 Anemoi contributors.
```

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        return {k: _detach_tensors(v) for k, v in obj.items()}
+    if isinstance(obj, (list, tuple)):
+        return type(obj)(_detach_tensors(v) for v in obj)
+    return obj
+
+
+def _sanitize_state(state: State) -> State:
+    """Remove private keys and convert tensors so the state is safe to pickle."""
+    clean = {k: v for k, v in state.items() if not k.startswith("_")}
+    return _detach_tensors(clean)
```
HCookie: I'm guessing that this is to remove the `_grib_templates_for_output` which cannot be pickled?
This may do some funky things when encoding, but not sure of a way around it. It also over eagerly drops the `_geography` key which is safe to pickle.

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+
+    chunks: list[State] = []
+    for i in range(num_chunks):
+        start = i * fields_per_chunk
+        end = min(start + fields_per_chunk, len(items))
+        if start >= len(items):
+            break
+        chunk = state.copy()
+        chunk["fields"] = dict(items[start:end])
+        chunks.append(chunk)
```
HCookie: Using itertools.islice might make things cleaner here?
`from itertools import islice`

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        metadata : Metadata
+            Metadata for the dataset.
+        output : Output | Any
+            The inner output (or its config dict) that will be forked into
+            writer processes.
+        num_writers : int, optional
+            Number of writer processes to spawn.
+            If not specified, an error is raised.
+        **kwargs : Any
+            Extra keyword arguments forwarded to :class:`ForwardOutput`.
```
HCookie: No longer subclassing from the `ForwardOutput`

```suggestion
            Extra keyword arguments forwarded to :class:`Output`.
```

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        super().__init__(context, metadata, **kwargs)
+
+        assert num_writers is not None, "Error. ParallelOutput selected but no 'num_writers' specified in config."
+        self.num_writers = num_writers
```
HCookie:
```suggestion
        num_writers: int,
        **kwargs: Any,
    ):
        """Initialise the ParallelOutput.

        Parameters
        ----------
        context : Context
            The inference context.
        metadata : Metadata
            Metadata for the dataset.
        output : Output | Any
            The inner output (or its config dict) that will be forked into
            writer processes.
        num_writers : int
            Number of writer processes to spawn.
        **kwargs : Any
            Extra keyword arguments forwarded to :class:`ForwardOutput`.
        """
        super().__init__(context, metadata, **kwargs)

        self.num_writers = num_writers
```
We can just not set a default if cannot be None

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+        """Write the state, dispatching to writer processes when enabled."""
+        step = state["step"]
+        if self.output_frequency is not None:
+            if (step % self.output_frequency).total_seconds() != 0:
+                return
+
+        if self.num_writers == 0:
+            # write in the main process without parallelism
+            return super().write_state(state)
+
```
HCookie: I think this will call `write_step` in the super class which will then trigger a `ValueError`.

src/anemoi/inference/outputs/parallel.py
HCookie: This could be a generator to then be used in the for loop below?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,308 @@
...
+            # Otherwise, post process and write the received state using the wrapped output
+            try:
+                message = self.post_process(message)
+
+                if not has_written_initial_state:
+                    # Some outputs (e.g. netcdf) must be explictly opened.
+                    if hasattr(output, "open"):
+                        LOG.debug("Writer %d openning %s", writer_id, output)
+                        output.open(message)
+                    # each process needs to write initial state to do some setup
```
HCookie: `open` is implemented as a dummy function on the top level, so can be called on all outputs.

```suggestion
                    output.open(message)
                    # each process needs to write initial state to do some setup
```

HCookie (changes requested):
Nice work on the parallel implementation, I've got a couple comments related to the use of some of the inference functions, and some comments on things I think can be cleaned up.

src/anemoi/inference/outputs/parallel.py
cathalobrien: good point

docs/inference/parallel_output.rst
gmertes: image

src/anemoi/inference/outputs/parallel.py
cathalobrien: You're right that _geography can be pickled.

`_grib_templates_for_output` and `_input` must be removed. not sure how to get around it. I could add back `_grib_templates_for_output` on the writer process side. but not sure how to do this cleanly

src/anemoi/inference/outputs/gribfile.py
cathalobrien: done

src/anemoi/inference/outputs/parallel.py
cathalobrien: you're right. I now check if num_writers >= 1

src/anemoi/inference/outputs/parallel.py
cathalobrien: sure, instead of making a generator I changed the function to return a state chunk given the writer id, but it no longer builds the full list

src/anemoi/inference/outputs/parallel.py
cathalobrien: done
```
                # read message type to determine how to process the message
                LOG.debug("Writer %d received '%s'", writer_id, message_type)
                if message_type == MessageType.TERMINATE:
                    output.close()
                    break
                elif message_type == MessageType.INITIAL_STATE:
                    output.write_initial_state(message)
                elif message_type == MessageType.STATE:
                    LOG.debug("Writer %d writing: %s", writer_id, message.get("date"))
                    output.write_state(message)
                    LOG.debug("Writer %d done: %s", writer_id, message.get("date"))
                else:
                    LOG.warning("Writer %d received message with unknown type '%s'", writer_id, message_type)
```

docs/inference/parallel_output.rst
cathalobrien: done

docs/inference/parallel_output.rst
```diff
@@ -0,0 +1,172 @@
...
+***********************
+ Troubleshooting
+***********************
+
+Writer process crashes
+=======================
+
+If a writer process crashes (e.g. due to a disk-full error or a bug in
+the output backend), the main process will detect that the writer is
+dead and abort inerence with a RuntimeError.
```
HCookie:
```suggestion
dead and abort inference with a RuntimeError.
```

src/anemoi/inference/decorators.py
```diff
@@ -84,6 +84,45 @@ def __init__(wrapped_cls, *args, **kwargs) -> None:
         return type(cls.__name__, (WrappedClass,), {})
 
 
+class supports_parallel_output:
+    """Decorator to indicate that an output supports parallel output.
+
+    Intercepts a ``suffix`` kwarg (injected by ParallelOutput) and
```
HCookie: Can we use a more special key, one that reasonably cannot be added by the user in the config, and is specific to the parallel output?
i.e. `parallel-output-suffix`?

src/anemoi/inference/decorators.py
```diff
@@ -84,6 +84,45 @@ def __init__(wrapped_cls, *args, **kwargs) -> None:
...
+    def __call__(self, cls: F) -> F:
+        # if not isinstance(cls, type):
+        #    raise TypeError(f"`{self.__class__.__name__}` can only be used to decorate classes")
+
+        class WrappedClass(cls):
+            def __init__(wrapped_cls, *args: Any, **kwargs: Any) -> None:
+                suffix = kwargs.pop("suffix", None)
+
+                if suffix and self.arg in kwargs and kwargs[self.arg] is not None:
+                    path = Path(kwargs[self.arg])
```
HCookie: We might want to check if it is str, and if not raise an error?

HCookie (commented):
Thanks for addressing my comments

src/anemoi/inference/outputs/parallel.py
gmertes: I think for now it's okay to accept the limitation that the input template provider will not work, either as a warning in the log or in the docs. It won't do any funky things, worst case it will just not find a template and crash.

cathalobrien: ok, I will add a warning when they are removed

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,317 @@
...
+        Each writer process runs this loop, which listens for incoming messages on its queue.
+        Messages are a tuple of (content, message_type), where 'message_type' indicates the type of message (e.g., MessageType.STATE, MessageType.INITIAL_STATE, MessageType.TERMINATE)
+        and 'content' is the state dictionary to write (for STATE and INITIAL_STATE messages).
+
+        Writers create their own output instance inside the writer process.
+
+        """
+        LOG.info("Writer %d started", writer_id)
+
+        output = create_output(context, output_config, self.metadata, parallel_output_suffix=f"_w{writer_id}", **kwargs)
```
HCookie: I do legitimately mean `parallel-output-suffix`, would prevent anyone from ever manually setting it

```suggestion
        output = create_output(context, output_config, self.metadata, **kwargs, **{'parallel-output-suffix'=f"_w{writer_id}",)
```

cathalobrien: haha ok sure

cathalobrien: done, I added an underscore to the front for even more obscurity

cathalobrien:
@copilot+claude-opus-4.6 please review this PR

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,325 @@
...
+                break
+
+            try:
+                # read message type to determine how to process the message
+                LOG.debug("Writer %d received '%s'", writer_id, message_type)
+                if message_type == MessageType.TERMINATE:
+                    output.close()
+                    break
+                elif message_type == MessageType.INITIAL_STATE:
+                    output.write_initial_state(message)
```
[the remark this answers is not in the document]
cathalobrien: added a new messageType.Open message to handle this

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,325 @@
...
+        calls 'write_state' on the wrapped output.
+        """
+        self._processes: list[mp.Process] = []
+        self._queues: list[mp.Queue] = []
+
+        for _ in range(self.num_writers):
+            # Use a bounded queue to prevent unlimited memory growth if the writers can't keep up with the main process
+            self._queues.append(mp.Queue(maxsize=10))
+
+        mp.set_start_method("fork", force=True)
```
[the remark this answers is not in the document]
cathalobrien: done

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,325 @@
...
+                queue.cancel_join_thread()
+            raise RuntimeError(
+                f"Writer {writer_id} is dead, inference will now fail. Check previous logs for errors in the writer process."
+            )
+
+    def write_state(self, state: State, message=MessageType.STATE) -> None:
+        """Write the state, dispatching to writer processes when enabled."""
+        for i in range(self.num_writers):
+            self._check_writer_alive(i)
+            chunk = _get_state_chunk(state, self.num_writers, i)
```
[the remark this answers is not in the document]
cathalobrien: Each writer applies post process on their chunks in parallel

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,351 @@
...
+                    i,
+                )
+                process.terminate()
+                process.join()
+                forcibly_teminated = True
+
+        LOG.info("ParallelOutput: all writers terminated.")
+        if forcibly_teminated:
+            raise RuntimeError(
+                "One or more writers were forcefully terminated after exceeding the shutdown timeout of %s. This can result in forecast data loss.Please check the logs for more details. Consider increasing the number of writer processes to avoid this.",
```
HCookie:
```suggestion
                "One or more writers were forcefully terminated after exceeding the shutdown timeout of %s. This can result in forecast data loss. Please check the logs for more details. Consider increasing the number of writer processes to avoid this.",
```

src/anemoi/inference/outputs/parallel.py
```diff
@@ -0,0 +1,351 @@
...
+        Messages are a tuple of (content, message_type), where 'message_type' indicates the type of message (e.g., MessageType.STATE, MessageType.INITIAL_STATE, MessageType.TERMINATE)
+        and 'content' is the state dictionary to write (for STATE and INITIAL_STATE messages).
+
+        Writers create their own output instance inside the writer process.
+
+        """
+        LOG.info("Writer %d started", writer_id)
+
+        output = create_output(
+            context, output_config, self.metadata, **kwargs, **{"_parallel-output-suffix": f"_w{writer_id}"}
```
HCookie, line 276: Could we make this a constant value? So it can be updated in one spot for both uses?

HCookie (dismissed):
Thanks for your work on this, 
Two minor comments, but I'd be happy to accept it as this.

src/anemoi/inference/outputs/parallel.py
cathalobrien: good idea, done
```
#src/anemoi/inference/decorators.py
PARALLEL_OUTPUT_SUFFIX_KWARG = "_parallel-output-suffix"
```

HCookie (approved):
Amazing work on this,
Apologies for the late approval, (I'd thought we'd merged this already....)

PR #506, opened 2026-05-29
feat: warn if patch_metadata adds new keys

### LINKED ISSUE #452: Warn users if metadata patches are updating non-existing keys
The `patch_metadata` entry in the inference configuration allows users to patch checkpoint metadata on the fly when running inference. These patches are specific for a certain metadata schema that depends on the version of anemoi packages used when creating the checkpoint. If one tries to apply the same patch to a checkpoint that has a new metadata schema, the patch is not applied as expected (instead of updating existing keys, it will add new ones) and this leads to problems that can be hard to debug. 

To address this, an option could be to raise a warning whenever _new_ metadata keys are created, and indicate which ones.

frazane:
## Description

The `patch_metadata` config entry patches checkpoint metadata on the fly. Patches are written against the metadata schema of the anemoi versions that produced the checkpoint. When the same patch is applied to a checkpoint with a **newer** metadata schema, the inner `merge()` in `Metadata.patch()` silently **adds** new keys instead of updating the intended existing ones — the update is lost and the failure is hard to debug.

## Changes

- `Metadata.patch()` now returns the dotted paths of metadata keys it **created** (only the top-most new key of any newly-created subtree is reported). The merge behavior is otherwise unchanged.
- The user-facing `patch_metadata` path in `Checkpoint` emits a `LOG.warning` naming those keys, so a schema mismatch is surfaced instead of failing silently.
- Internal patches in `external_graph.py` (which intentionally add keys) simply ignore the return value, so they produce no spurious warnings.
- Adds unit tests for the returned key paths and for the warning.

Closes #452

### DIFF
--- a/src/anemoi/inference/checkpoint.py
+++ b/src/anemoi/inference/checkpoint.py
@@ -181,7 +181,16 @@ class Checkpoint:
             for dataset, metadata in multi_metadata.items():
                 patch = multi_datasets_config(self.patch_metadata, dataset, list(multi_metadata.keys()), strict=False)
                 LOG.warning(f"[{dataset}] Patching metadata with {patch}")
-                metadata.patch(patch)
+                new_keys = metadata.patch(patch)
+                if new_keys:
+                    LOG.warning(
+                        "[%s] The metadata patch created %d new key(s) that did not exist before: %s. "
+                        "This usually means the patch does not match this checkpoint's metadata schema, "
+                        "so the intended update was not applied.",
+                        dataset,
+                        len(new_keys),
+                        ", ".join(new_keys),
+                    )
 
         return multi_metadata
 
--- a/src/anemoi/inference/checkpoint.py
+++ b/src/anemoi/inference/checkpoint.py
@@ -184,9 +184,9 @@ class Checkpoint:
                 new_keys = metadata.patch(patch)
                 if new_keys:
                     LOG.warning(
-                        "[%s] The metadata patch created %d new key(s) that did not exist before: %s. "
-                        "This usually means the patch does not match this checkpoint's metadata schema, "
-                        "so the intended update was not applied.",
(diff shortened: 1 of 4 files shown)


gmertes (commented):
Thanks for implementing this, and the overall implementation LGTM!

There is one case where adding a new key is expected, and that is the dataset [constant_fields](https://github.com/ecmwf/anemoi-inference/blob/9cc8adef94b40288b0da86171f874414879720c3/src/anemoi/inference/metadata.py#L442-L448) patch. That entry is sometimes missing and we often want to add it back in. 

You do say in the code and in your warning that sometimes adding news keys is intentional, but does it still make sense to show a warning for cases that we know are intentional? I'm worried that could be a bit confusing.

The patch does become a bit clunky in multi-dataset world because you have to write it as:

```yaml
patch_metadata:
  dataset:
    data:  # or whatever name you have
      constant_fields: [z, sdor, slor, lsm]
```

We could consider making this a dedicated config entry, rather than going through patch_metadata. But maybe for now it could be worth adding a warning exclusion for constant_fields?

frazane:
Hi Gert, thanks! I agree that we can add a warning exclusion for now. However I see that this `constant_fields` patch is not in the actual metadata, it's only a proxy or "syntax sugar" for `config.dataset.<dataset-name>.variables_metadata.<variable-name>.constant_in_time` and kind of breaks the idea that we pass actual metadata keys to `patch_metadata`...should we not just remove it? 

I would not call the patch particularly clunky...all our patches look like: https://github.com/MeteoSwiss/evalml/blob/main/resources/inference/metadata/sgm-multidataset-ich1-oper-patch.yaml 😂

gmertes (approved):
Yes I fully agree! But removing could break people's setups. So we need to think about backwards compatibility and/or do a deprecation period. Are you happy to merge the PR like this for now?

PR #523, opened 2026-06-22
chore: Update to earthkit 1.0

aaron-hopkinson:
Update earthkit-data dep to >=1.0.

### DIFF
--- a/src/anemoi/inference/grib/encoding.py
+++ b/src/anemoi/inference/grib/encoding.py
@@ -304,7 +304,7 @@ def grib_keys(
 
     # 1 if local definition is present, like for ECMWF GRIBs
     if template is not None:
-        local_use_present = template.metadata("localUsePresent", default=0)
+        local_use_present = template.get("metadata.localUsePresent", default=0)
     else:
         local_use_present = 0
 
@@ -399,7 +399,7 @@ def check_encoding(handle: Any, keys: dict[str, Any], first: bool = True) -> Non
 
         if first:
             import eccodes
-            from earthkit.data.readers.grib.codes import GribCodesHandle
+            from earthkit.data.readers.grib.handle import GribCodesHandle
 
             handle = GribCodesHandle(eccodes.codes_clone(handle._handle), None, None)
             return check_encoding(handle, keys, first=False)
@@ -436,7 +436,15 @@ def encode_message(
         The encoded GRIB handle.
     """
     metadata = metadata.copy()  # avoid modifying the original metadata
-    handle = template.handle.clone()
+    if hasattr(template, 'handle'):
+        # Dummy wrapper or legacy object with handle attribute
+        handle = template.handle.clone()
+    else:
+        # New earthkit Field — get handle via private GRIB metadata
+        grib = template._get_grib(strict=True)
+        if grib is None:
+            raise ValueError("Template is not a GRIB field, cannot encode message")
+        handle = grib.handle.clone()
 
     if check_nans and values is not None:
         import numpy as np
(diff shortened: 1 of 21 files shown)


src/anemoi/inference/grib/encoding.py
```diff
@@ -436,7 +436,15 @@ def encode_message(
...
-    handle = template.handle.clone()
+    if hasattr(template, "handle"):
+        # Dummy wrapper or legacy object with handle attribute
+        handle = template.handle.clone()
+    else:
+        # New earthkit Field — get handle via private GRIB metadata
+        grib = template._get_grib(strict=True)
+        if grib is None:
+            raise ValueError("Template is not a GRIB field, cannot encode message")
+        handle = grib.handle.clone()
```
gmertes, line 447: This logic needs a check, it's not clear in which situation the legacy handler is needed.

src/anemoi/inference/inputs/cds.py
```diff
@@ -99,7 +99,7 @@ def _(r: DataRequest) -> str:
 
         LOG.debug("%s", _(r))
 
-        result += ekd.from_source("cds", d, r)
+        result = concat(result, ekd.from_source("cds", d, r).to_fieldlist())
```
gmertes, line 102: `concat` is not imported?

src/anemoi/inference/grib/encoding.py
```diff
@@ -304,7 +304,7 @@ def grib_keys(
 
     # 1 if local definition is present, like for ECMWF GRIBs
     if template is not None:
-        local_use_present = template.metadata("localUsePresent", default=0)
+        local_use_present = template.get("metadata.localUsePresent", default=0)
```
gmertes, line 307: This is a grib-only key, is this the correct way to access those?

src/anemoi/inference/outputs/gribfile.py
```diff
@@ -222,12 +222,13 @@ def write_message(self, message: FloatArray, template: ekd.Field, **keys: dict[s
...
         # Make sure `name` is not in the keys, otherwise grib_encoding will fail
-        if template is not None and template.metadata("name", default=None) is not None:
+        if template is not None and template.get("labels.name", default=None) is not None:
             # We cannot clear the metadata...
             class Dummy:
                 def __init__(self, template: ekd.Field) -> None:
                     self.template = template
-                    self.handle = template.handle
+                    grib = template._get_grib(strict=True)
+                    self.handle = grib.handle if grib is not None else None
```
gmertes, line 231: I don't think handle can be None here for the downstream usage of this class. Template will always be a Field created from grib, so there should always be a handle.

src/anemoi/inference/inputs/ekd.py
```diff
@@ -480,25 +548,18 @@ def set_private_attributes(self, state: State, fields: ekd.FieldList) -> None:
...
+        if grid := get_geography_info("grid"):
             geography_information["grid"] = grid
 
-        else:  # If grid is undefined we don't want to add the area
-            if grid:
-                geography_information["grid"] = grid
-
-            if area := get_geography_info("mars_area"):
-                geography_information["area"] = area
-
```
gmertes, line 501: @HCookie TODO: check this logic

src/anemoi/inference/grib/encoding.py
```diff
@@ -436,7 +436,15 @@ def encode_message(
         The encoded GRIB handle.
     """
     metadata = metadata.copy()  # avoid modifying the original metadata
-    handle = template.handle.clone()
+    if hasattr(template, "handle"):
+        # Dummy wrapper or legacy object with handle attribute
+        handle = template.handle.clone()
+    else:
+        # New earthkit Field — get handle via private GRIB metadata
+        grib = template._get_grib(strict=True)
```
gmertes, line 444: Check if we can get a public method to get the handle.

gmertes (commented):
Quick first pass. I think the logic in `inputs/ekd.py` and `post_pocessors/earthkit_state.py` needs the most scrutiny, everything else seems straightforward.

PR #526, opened 2026-06-23
feat(outputs/grib): allow configuration of grib missing_value sentinel

cosunae:
## Description
Adds a configurable `missing_value` parameter that flows from the inference YAML config
through to GRIB encoding. This fixes a silent data corruption bug where physically valid
geophysical values could collide with the hardcoded `missingValue=9999` sentinel used
during GRIB bitmap encoding. In mountainous or elevated regions, a geopotential value of 9999 is a physical value. 

**Bug fix** — silent missing-value collision in GRIB output encoding.

When writing GRIB output with NaN masking enabled (`allow_nans: true`), anemoi-inference
replaces NaN values with a sentinel before calling eccodes `set_values()`. The sentinel
must not collide with any physically valid value in the field being encoded.

The previous hardcoded sentinel of `9999` caused incorrect bitmap encoding for
geopotential at 850 hPa (`z_850`) over the Swiss Alps: the geopotential height at
~1019 m altitude yields a geopotential value of ~9999 m²/s², which after GRIB
quantization rounds to exactly `9999`. eccodes then interpreted this real grid point
as missing (`bitmap=0`), producing a spurious NaN in the decoded output.

The fix changes the default sentinel to `9.999e20` — a value that cannot appear as a
physically valid geophysical quantity — and makes it user-configurable via the
inference config so that it can be matched to any custom GRIB template if needed.

It is a bug fix, so there is no related issue.

##  Additional notes ##
**On the GRIB round-trip**: `missingValue` is not stored in the GRIB2 binary — it is
resolved at read time from eccodes concept definitions (where is the default is `9999`). This means
the sentinel used during encoding and the one seen on decode are different
(`9.999e20` vs `9999`). I assume the harcoded default value here it was chosen 9999, and I considered whether this would break functionality when decoding the data using eccodes definitions that still contain a default missing value of 9999. However, I think it is safe because inference encode explicitly a bitmap section in the grib message. In that case, it should get precedence and the missing value of eccodes definitions (in presence of a bitmap) would be ignored.

**Breaking change consideration**: The default `missing_value` changes from `9999` to
`9.999e20`. Any downstream consumer that detects missing values by comparing against
the sentinel value (rather than reading the bitmap) will now see `9.999e20` instead of
`9999` for NaN-masked points. The correct approach is always to use the bitmap; GRIB2
consumers using earthkit.data or standard eccodes `get_values` would be unaffected. Other readers might get affected if not using the bitmap.

BEGIN_COMMIT_OVERRIDE
feat(grib output): allow configuration of missing_value sentinel (#526)
END_COMMIT_OVERRIDE

### DIFF
--- a/src/anemoi/inference/config/run.py
+++ b/src/anemoi/inference/config/run.py
@@ -83,7 +83,7 @@ class RunConfiguration(Configuration):
     missing_value: int | float = 9.999e20
     """The value used to replace NaN values when encoding GRIB output (requires allow_nans=True).
     Using a large sentinel (default 9.999e20) avoids collision with physically valid
-    geophysical values (e.g. z_850 ~ 9999 m2/s2 over Switzerland).
+    geophysical values (e.g. z_850).
     """
 
     use_grib_paramid: bool = False
--- a/src/anemoi/inference/config/run.py
+++ b/src/anemoi/inference/config/run.py
@@ -80,12 +80,6 @@ class RunConfiguration(Configuration):
     - If True, the model will allow NaNs in the input and output.
     """
 
-    missing_value: int | float = 9.999e20
-    """The value used to replace NaN values when encoding GRIB output (requires allow_nans=True).
-    Using a large sentinel (default 9.999e20) avoids collision with physically valid
-    geophysical values (e.g. z_850).
-    """
-
     use_grib_paramid: bool = False
     """If True, the runner will use the grib parameter ID when generating MARS requests."""
 
--- a/src/anemoi/inference/config/run.py
+++ b/src/anemoi/inference/config/run.py
@@ -82,7 +82,6 @@ class RunConfiguration(Configuration):
 
     missing_value: int | float = 9.999e20
     """The value used to replace NaN values when encoding GRIB output (requires allow_nans=True).
-    Must match the missingValue set in the GRIB templates used for output.
     Using a large sentinel (default 9.999e20) avoids collision with physically valid
     geophysical values (e.g. z_850 ~ 9999 m2/s2 over Switzerland).
     """
--- a/src/anemoi/inference/config/run.py
+++ b/src/anemoi/inference/config/run.py
@@ -80,6 +80,13 @@ class RunConfiguration(Configuration):
     - If True, the model will allow NaNs in the input and output.
     """
 
+    missing_value: int | float = 9.999e20
+    """The value used to replace NaN values when encoding GRIB output (requires allow_nans=True).
+    Must match the missingValue set in the GRIB templates used for output.
+    Using a large sentinel (default 9.999e20) avoids collision with physically valid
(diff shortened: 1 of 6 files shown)


src/anemoi/inference/config/run.py
```diff
@@ -80,6 +80,13 @@ class RunConfiguration(Configuration):
     - If True, the model will allow NaNs in the input and output.
     """
 
+    missing_value: int | float = 9.999e20
+    """The value used to replace NaN values when encoding GRIB output (requires allow_nans=True).
+    Must match the missingValue set in the GRIB templates used for output.
+    Using a large sentinel (default 9.999e20) avoids collision with physically valid
+    geophysical values (e.g. z_850 ~ 9999 m2/s2 over Switzerland).
+    """
```
HCookie: I think it'd be best to put this in the grib output init block, and configure it there, not in the top level.
Then you wont have to carry it so far.

src/anemoi/inference/grib/encoding.py
```diff
@@ -413,7 +413,7 @@ def encode_message(
     template: Any,
     metadata: dict[str, Any],
     check_nans: bool = False,
-    missing_value: int | float = 9999,
+    missing_value: int | float = 9.999e20,
```
HCookie: -9999 is probably a better value, an e20 runs the risk of causing other encoding issues, particularly if the range is too big, small differences will be lost.

HCookie (changes requested):
Hi,
Thanks for this PR, I have a couple comments and think it'd be best to put this directly in the grib output block.

src/anemoi/inference/grib/encoding.py
cosunae: I am not sure I understand what other issues e20 could raise. It is difficult to think what values are generally unphysical, and while -9999 a very large value like e20 sounded more bullet proof. 
But I think -9999 should also work fine for all variables, or at least I hope so

src/anemoi/inference/config/run.py
cosunae: makes sense for sure, I was not too familiar with the code base. Can you check is this how you meant it? https://github.com/ecmwf/anemoi-inference/pull/526/changes/8e1aa0d6771e5244a683681ce2d2383abed27781

src/anemoi/inference/grib/encoding.py
HCookie: Let's set it as -9999 for now, e20 is overkill for most things, but can remain configurable

src/anemoi/inference/config/run.py
HCookie: Yep, thanks for making that change, exactly correct.

HCookie (approved):
Thanks for contributing this feature.
LGTM

PR #530, opened 2026-06-24
feat(dataset input): add use_trajectories option

### LINKED ISSUE #529: Running inference with trajectory/forecast forcing, i.e. for ocean or land models.

The ocean and land are hugely influenced by the atmospheric forcing. As such, the meaningful performance test, is how they run when forced with fields from an atmospheric forecast. The trajectory updates to anemoi datasets means its now possible to create a dataset which has multiple forecasts, i.e. a 10-day lead time forecast for each date within the dataset, but its not possible to use this to force the model during inference

I would like to be able to point to a dataset which contains forecast trajectories, and to run inference using the forecast trejctory. So, for 1st Jan start date, the model takes 1st jan initial conditions from the dataset, and then takes atmospheric forcing for lead times 0-240, all from date 1st jan, and uses these to force a 10 day run. (currently, the set up takes 1st jan initial conditions, and then takes forcing from 2nd, 3rd, ...10th jan. i.e. analysis forcing valid for the correct day, not forecast forcing)

_No response_

### Additional context

_No response_

### Organisation

ECMWF

HCookie:
Adds `use_trajectories` support to `DatasetInput`, allowing trajectory datasets (5D: base_dates, variables, ensembles, steps, cells) to be used as inputs.

Resolves #529 
- [x] Validated on a real checkpoint and ds

```yaml
checkpoint: CKPT
date: 2023-01-02

input: validation

dynamic_forcings:
  dataset: 
    dataset: Trajectory.zarr
    use_trajectories: true
```

### Changes
- `DatasetInput` accepts `use_trajectories=True` to enable trajectory mode
- `_load_trajectories` loads data by indexing into the step dimension relative to a base date
- Uses `step_frequency` and `base_dates` (appropriate for trajectory datasets) instead of `frequency` and `dates`
- Transposes trajectory data to match the expected `(dates, variables, ensemble, cells)` shape

----
📚 Documentation preview 📚: https://anemoi-inference--530.org.readthedocs.build/en/530/

### DIFF
--- a/docs/inference/configs/inputs.rst
+++ b/docs/inference/configs/inputs.rst
@@ -50,6 +50,9 @@ This allows for forecasts to be included in the dataset, and for the model to in
 
 Setting `use_trajectories` in a dataset input will allow for the model to be initialised from the forecast, filtering by valid_time, rather than just the basetime.
 
+In the case of the initial conditions, the prior dates, the model will be initialised from step 0 of the dataset, i.e. the analysis, with future forcings coming from the forecast steps, with the base date
+being the date specified in the config, the base_date.
+
 
 .. code:: yaml
 
--- a/docs/inference/configs/inputs.rst
+++ b/docs/inference/configs/inputs.rst
@@ -46,9 +46,9 @@ Trajectory Inputs
 #################
 
 `anemoi-datasets` `v0.5.38` introduced a new feature to allow for trajectory datasets, introducing a 5th dimension to the dataset.
-This allows for forecasts to be included in the dataset, and for the model to initialised from steps of a model, not just the analysis.
+This allows for forecasts to be included in the dataset, and for inference it means that it is possible for the model to be forced from steps of a model, not just the analysis.
 
-Setting `use_trajectories` in a dataset input will allow for the model to be initialised from the forecast, filtering by valid_time, rather than just the basetime.
+Setting `use_trajectories` in a dataset input will allow for the model to be forced from the forecasts within the dataset, filtering by step and the basetime of the run, rather than just the absolute time throughout the run.
 
 In the case of the initial conditions, the prior dates, the model will be initialised from step 0 of the dataset, i.e. the analysis, with future forcings coming from the forecast steps, with the base date
 being the date specified in the config, the base_date.
--- a/docs/inference/configs/inputs.rst
+++ b/docs/inference/configs/inputs.rst
@@ -41,6 +41,22 @@ You can also provide a full dataset specification as follows:
 See :ref:`anemoi-datasets:opening-datasets` in the documentation of the
 `anemoi-datasets` package for more information on how to open datasets.
 
+#################
+Trajectory Inputs
+#################
+
+`anemoi-datasets` `v0.5.38` introduced a new feature to allow for trajectory datasets, introducing a 5th dimension to the dataset.
+This allows for forecasts to be included in the dataset, and for the model to initialised from steps of a model, not just the analysis.
+
+Setting `use_trajectories` in a dataset input will allow for the model to be initialised from the forecast, filtering by valid_time, rather than just the basetime.
+
+
+.. code:: yaml
+
+   input:
+      test:
+       use_trajectories: true
+
 ******
  grib
 ******
--- a/docs/inference/configs/inputs.rst
+++ b/docs/inference/configs/inputs.rst
@@ -54,8 +54,9 @@ Setting `use_trajectories` in a dataset input will allow for the model to be ini
 .. code:: yaml
 
    input:
-      test:
-       use_trajectories: true
+      dataset:
+         PATH_GOES_HERE
+         use_trajectories: true
 
 ******
  grib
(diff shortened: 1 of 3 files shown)


docs/inference/configs/inputs.rst
```diff
@@ -41,6 +41,22 @@ You can also provide a full dataset specification as follows:
...
+This allows for forecasts to be included in the dataset, and for the model to initialised from steps of a model, not just the analysis.
+
+Setting `use_trajectories` in a dataset input will allow for the model to be initialised from the forecast, filtering by valid_time, rather than just the basetime.
+
+
+.. code:: yaml
+
+   input:
+      test:
+       use_trajectories: true
```
mc4117: I think it would make more sense to give the example with grib?

src/anemoi/inference/inputs/dataset.py
```diff
@@ -248,6 +282,57 @@ def _load_dates(self, dates: list[Date]) -> Any:
...
+            List of dates for which to load the data.
+        base_date : Any
+            The base date for relative date calculations.
+
+        Returns
+        -------
+        Any
+            The loaded data.
+        """
+        base_idx = self._find_index_for_date(base_date)
```
mc4117, line 316: Can you specify which steps you want to use from the trajectory dataset?

docs/inference/configs/inputs.rst
HCookie: It's not relevant for the grib input? This is for datasets...?

src/anemoi/inference/inputs/dataset.py
HCookie: Wdym? Would it not be set by the dates coming in?

docs/inference/configs/inputs.rst
mc4117: sorry I meant zarr

gmertes (commented):
The refactoring of the date logic seems significant. Is there any way we can cover with tests that the old behaviour isn't changing for the non-trajectory case? Do the integration tests already capture this?

HCookie:
@gmertes While it does change a bit, the entire old pathway was simply a move of the function, and slight functional separation. I'll look into adding tests. As anemoi-datasets tho is tricky to mock, this may be tricky to do.

gmertes (dismissed):

gmertes:
Would be good to get @mc4117 or @RachelFurner to test before we merge?

RachelFurner:
> Would be good to get @mc4117 or @RachelFurner to test before we merge?

I've taken a look and run a simple test case with standard and trajectory-forecast forcing. It looks like the code does what I need.

RachelFurner (commented):
Looking through the code, for me this phrasing in the inputs.rst feels confusing

    "This allows for forecasts to be included in the dataset, and for the model to initialised from steps of a model, not just the analysis.
    
    Setting `use_trajectories` in a dataset input will allow for the model to be initialised from the forecast, filtering by valid_time, rather than just the basetime.
    
    In the case of the initial conditions, the prior dates, the model will be initialised from step 0 of the dataset, i.e. the analysis, with future forcings coming from the forecast steps, with the base date
    being the date specified in the config, the base_date."

I'm not sure that there is a use case where one would want to *initialise* from trajectories -- my use case, and I imagine most, would be around *forcing* from trajectories. Can the forcing (and potentially initialisation, and more generically 'inputs') use cases be clarified a little here please.

docs/inference/configs/inputs.rst
```diff
@@ -41,6 +41,26 @@ You can also provide a full dataset specification as follows:
...
+#################
+
+`anemoi-datasets` `v0.5.38` introduced a new feature to allow for trajectory datasets, introducing a 5th dimension to the dataset.
+This allows for forecasts to be included in the dataset, and for the model to initialised from steps of a model, not just the analysis.
+
+Setting `use_trajectories` in a dataset input will allow for the model to be initialised from the forecast, filtering by valid_time, rather than just the basetime.
+
+In the case of the initial conditions, the prior dates, the model will be initialised from step 0 of the dataset, i.e. the analysis, with future forcings coming from the forecast steps, with the base date
+being the date specified in the config, the base_date.
+
```
HCookie, line 55:
```suggestion
`anemoi-datasets` `v0.5.38` introduced a new feature to allow for trajectory datasets, introducing a 5th dimension to the dataset.
This allows for forecasts to be included in the dataset, and for inference it means that it is possible for the model to be forced from steps of a model, not just the analysis.

Setting `use_trajectories` in a dataset input will allow for the model to be forced from the forecasts within the dataset, filtering by step and the basetime of the run, rather than just the absolute time throughout the run.

In the case of the initial conditions, the prior dates, the model will be initialised from step 0 of the dataset, i.e. the analysis, with future forcings coming from the forecast steps, with the base date
being the date specified in the config, the base_date.

```
@RachelFurner thoughts?

RachelFurner: Line 49 "model to *be* forced", otherwise fabulous, thanks!

gmertes (approved):

PR #537, opened 2026-07-03
fix(parallel output): Support parallel grib archive requests

gmertes:
## Description
For parallel grib output, we forgot about the archive requests. These also need to have the suffix attached so a unique file is created by each writer.

Update the `supports_parallel_output` decorator so that it:

1. Can take multiple arguments, all of which will have the suffix attached
2. Supports dot notation for nested dictionaries, like `archive_requests.path`

Also rename the parallel unit test file to `test_parallel.py` for clarity.


HCookie (approved):
Whoopsies, looks good to me

PR #546, opened 2026-07-09
feat(post_processors): optionally emit zero-valued step 0 field in Accumulate

### LINKED ISSUE #545: `accumulate_from_start_of_forecast` does not emit step 0 in its output
The `Accumulate` post-processor (`accumulate_from_start_of_forecast`) correctly accumulates fields from zero, but never emits a step 0 field. The accumulator is initialised to zeros implicitly on the first forecast step, so the initial condition (accumulation = 0 at t=0) is never included in the output state.

As a consequence, downstream consumers that need to de-accumulate the output (i.e. convert cumulative-from-start to per-step accumulations via differencing) require step 0 as a reference point. Without it they must either:

- synthesise a zero-valued step 0 themselves (requiring knowledge of this implementation detail), or
- silently drop the first accumulation window.

The expected behaviour should be thta when `accumulate_from_start_of_forecast` is active, a step 0 field with value 0 everywhere should be present in the output, consistent with the physical meaning that no accumulation has occurred at the initial condition.

One possible solution could be to emit a zero-valued field at `step=0` for each variable in `self.accumulations` before the first `process()` call.

dnerini:
## Description
This updates the `Accumulate` post-processor to optionally emit zero-valued fields at step=0 as a reference for downstream differencing.

The `Accumulate` post-processor (`accumulate_from_start_of_forecast`) accumulates model output fields (e.g., precipitation) across forecast steps. Downstream systems that need to reverse the accumulation (via differencing) require a zero-valued reference at step=0.

**Previous behaviour:** When `process()` is called on the initial state (step=0), accumulation fields (like `tp`) are typically absent because they are model outputs, not initial conditions. The processor passes through unchanged, as no zero-valued field is added.

**New behavior:** The initial state output can optionally include a zero-valued field for each accumulation variable, representing "no accumulation has occurred yet at the initial condition."

Closes #545.

##  Additional notes ##

- `emit_initial_zeros=False` is the default to avoid breaking existing workflows that did not rely on the zero reference.
- Zero-valued fields use shape from `metadata.number_of_grid_points`
- If an accumulation field already exists in the initial state with a non-zero value, it is overridden to 0 when `emit_initial_zeros=True`.
- The accumulators (`self.accumulators`) are only populated from the second call onward, so the first real forecast step (step=1h, 6h, etc.) initializes them properly from zeros

This PR is AI-assisted.


src/anemoi/inference/post_processors/accumulate.py
```diff
@@ -75,6 +85,17 @@ def process(self, state: State) -> State:
         """
         state = state.copy()
         state.setdefault("start_steps", {})
+
+        if not self._initialized:
```
HCookie: I think adding a check here that this is writing the initial state is needed. Otherwise if `write_initial_state` is False, this will override on the first real step.

src/anemoi/inference/post_processors/accumulate.py
```diff
@@ -75,6 +85,17 @@ def process(self, state: State) -> State:
...
+            # Emit zero-valued fields at step=0 (initial state).
```
HCookie: Technically, this is only coincidentally true, not enforced.

src/anemoi/inference/post_processors/accumulate.py
```diff
@@ -75,6 +85,17 @@ def process(self, state: State) -> State:
...
+            # Physical meaning: no accumulation has occurred at the initial condition.
+            n_points = len(state["latitudes"])
```
HCookie: Would be best to either use another field, or get it from the grid_points attribute in the metadata.

src/anemoi/inference/post_processors/accumulate.py
```diff
@@ -75,6 +85,17 @@ def process(self, state: State) -> State:
...
+            for accumulation in self.accumulations:
+                state["fields"][accumulation] = np.zeros(n_points)
```
HCookie, line 110: I think this should check to ensure it's not in

HCookie (changes requested):

gmertes (commented):
Thanks for the PR Daniele! I think this is a nice feature to have, but I'd like to be a bit cautious. Not outputting any diagnostics at step 0, including accumulations, was done on purpose. For your workflow this may be a fix, for other people's workflows this may be breaking. 

Could we put this feature behind a flag that is turned off by default? Then later if we see that most people run with the flag turned on, we can consider changing the default value.

src/anemoi/inference/post_processors/accumulate.py
dnerini: see 4c2c60f

src/anemoi/inference/post_processors/accumulate.py
dnerini: see 4c2c60f

src/anemoi/inference/post_processors/accumulate.py
dnerini: see f8682d8

src/anemoi/inference/post_processors/accumulate.py
dnerini: looks like the comment is incomplete?

dnerini:
Thanks @HCookie and @gmertes for the review and apologies for taking so long to react to your comments.

@HCookie I address your comments above, can you please have another look?

@gmertes I introduce a flag to make this feature optional (off by default), what do you think?

gmertes (commented):
Thanks Daniele, this look good to me. Harrison is back from leave next week, shall we wait for him to check that last remaining comment?

dnerini:
Yes I'm happy to wait for his return. Did you check the discussion on the related issue? I wonder if we should change the encoding of the step=0 field accordingly?

dnerini:
After some more discussions, we propose to close this PR as it can't address the GRIB encoding issue raised in #545 (or it could, but it'd be a very large and ugly change for little value). 

Instead, the suggestion is to implement a bespoke plugin that can generate the step=0 zero fields with the right parameter and grib metadata.

PR #547, opened 2026-07-10
fix(parallel output): grib input templates

anaprietonem:
## Description
This PR is to update the ParallelOutput runner to correctly carry the GRIB templates to writer processes. 
- The templates are earthkit Field objects wrapping live handles, which can't be pickled and previously had to be dropped from the state. 
- Now, inside ``_sanitise_state``, each template is serialised to raw GRIB bytes and stored unde5``_grib_templates_bytes_for_output``;
-  Then the writer loop calls a new ``_restore_grib_templates``helper that turns those bytes back into a tiny ``_GribHandleWrapper`` and puts them back under the usual ``_grib_templates_for_output`` key before handing the state to the wrapped output. 
- All of this lives entirely in ``parallel.py`` so the rest of the codebase are unchanged from main and only the parallel path is affected.

Unit-tests are added too!

##  Additional notes ##

### DIFF
--- a/src/anemoi/inference/outputs/parallel.py
+++ b/src/anemoi/inference/outputs/parallel.py
@@ -49,24 +49,6 @@ def _detach_tensors(obj: Any) -> Any:
     return obj
 
 
-class _GribHandleWrapper:
-    """Minimal wrapper around a ``GribCodesHandle`` reconstructed from raw GRIB bytes.
-
-    Satisfies the ``template.metadata(...)`` and ``template.handle.clone()`` contract
-    expected by the GRIB encoder, so the writer processes in :class:`ParallelOutput`
-    can consume templates that were shipped across the process boundary as bytes.
-    """
-
-    def __init__(self, handle: Any) -> None:
-        self.handle = handle
-
-    def metadata(self, key: str, default: Any = None) -> Any:
-        try:
-            return self.handle.get(key, default=default)
-        except Exception:
-            return default
-
-
 def _serialise_grib_templates(templates: dict) -> dict[str, bytes]:
     """Convert a dict of earthkit GRIB fields to a dict of raw GRIB bytes so they can be pickled."""
     result = {}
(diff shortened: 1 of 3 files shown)


src/anemoi/inference/outputs/parallel.py
```diff
@@ -49,19 +49,93 @@ def _detach_tensors(obj: Any) -> Any:
...
 
     state = state.copy()
+
+    # Serialise GRIB templates to raw bytes so they survive pickling into writer processes.
+    # The writer reconstructs ekd.Field-like objects from these bytes via
+    # _restore_grib_templates() before handing the state to the wrapped output.
+    if (templates := state.get("_grib_templates_for_output")) is not None:
+        state["_grib_templates_bytes_for_output"] = _serialise_grib_templates(templates)
+        state.pop("_grib_templates_for_output")
+        LOG.debug("Serialised %d GRIB templates to bytes for writer process", len(templates))
```
HCookie: I'm slightly concerned about this and it's interactions with our wrapped earthkit objects. As we can use templates on the other side, can this warn if it fails to byte and continue?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -49,19 +49,93 @@ def _detach_tensors(obj: Any) -> Any:
...
+    passing the state to the wrapped output. Failed reconstructions are skipped
+    with a warning so a single bad template does not abort the whole state.
+    """
+    import eccodes
+    from earthkit.data.readers.grib.codes import GribCodesHandle
+
+    result: dict[str, _GribHandleWrapper] = {}
+    for name, msg in bytes_templates.items():
+        try:
+            h = eccodes.codes_new_from_message(msg)
```
HCookie: Could this use the earthkit.data.from_source('memory'), so we don't need to import eccodes?

src/anemoi/inference/outputs/parallel.py
```diff
@@ -49,19 +49,93 @@ def _detach_tensors(obj: Any) -> Any:
...
+        except Exception:
+            return default
+
+
+def _serialise_grib_templates(templates: dict) -> dict[str, bytes]:
+    """Convert a dict of earthkit GRIB fields to a dict of raw GRIB bytes so they can be pickled."""
+    result = {}
+    for name, field in templates.items():
+        try:
+            result[name] = field.message()
```
HCookie: That's going to be a big message including the data....

src/anemoi/inference/outputs/parallel.py
anaprietonem: can check!

src/anemoi/inference/outputs/parallel.py
anaprietonem: do you have an example or where this could problematic, or what you mean with using templates on the other side?

anaprietonem: And if we allow this to fail and continue, then the risk is that we get a result mis-match when using anemoi-inference parallel runner for writing vs the serial one

src/anemoi/inference/outputs/parallel.py
anaprietonem: fair, are you concerned that this could lead to larger memory footprint? @cathalobrien do you know if using the inference profiler could be possible here to check this?

HCookie: It might be slower due to the transfer cost

src/anemoi/inference/outputs/parallel.py
HCookie: When we do odd transform things, we get an almost earthkit field, and it might not behave as you expect. If we fail, warn and continue we can at least get output. Not all uses of the parallel output will write grib, so blocking on that is not a good idea. Plus your differences could just be an issue with the templates.

anaprietonem: Okey I might be missing context here but from the docs: https://anemoi.readthedocs.io/projects/inference/en/latest/inference/configs/grib-output.html#templates
```
Use the messages from the GRIB input as templates. Only works with a GRIB input.
```
So you're right not everyone using the parallel output will write grib, but the serialising would just fire if GRIB is used cause otherwise _grib_templates_for_output shouldn't be part of the state?

HCookie: But grib input might be used without a grib output

HCookie (approved):
Thanks for addressing our comments, looks good to me, lets wait for Gert's review tho too

gmertes (approved):
Thanks Ana!

PR #549, opened 2026-07-14
feat(netcdf output)!: Produce netCDF-files that follow CF conventions

mpvginde:
Currently the netCDF output produced by inference is quite minimal in terms of coordinate metadata.

image

Some issues: 
- latitude and longitude are not marked as coordinate variables
- one single time coordinate
  - may lead to loss of information if used together with `write_initial_state: false`. 
  - prevents use of `xr.open_mfdatasets()`
- generally no CF conventions are followed.

This PR updates the netCDF writing, bringing the netCDF-output in line with CF conventions. Which in term provides better `xarray`-handling

image

### DIFF
--- a/src/anemoi/inference/outputs/netcdf.py
+++ b/src/anemoi/inference/outputs/netcdf.py
@@ -10,6 +10,8 @@
 import logging
 import os
 import threading
+from datetime import datetime
+from datetime import timezone
 from pathlib import Path
 
 import numpy as np
@@ -31,6 +33,8 @@ LOG = logging.getLogger(__name__)
 
 # In case HDF5 was not compiled with thread safety on
 LOCK = threading.RLock()
+
+CALENDAR = "standard"
 
 
 @output_registry.register("netcdf")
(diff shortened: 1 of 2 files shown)


src/anemoi/inference/testing/checks.py
```diff
@@ -157,17 +157,13 @@ def check_with_xarray(
     if expected_time is not None:
         assert len(ds.time) == expected_time, f"Expected {expected_time} time steps, but found {len(ds.time)}."
 
-    # lat and lon are variables in the file, but they are not variables of the model so skip them
-    skip = {"latitude", "longitude"}
-    params = set(ds.data_vars) - skip
+    params = set(ds.data_vars)
```
gmertes: Including the lat/lons here is breaking the zarr-out integration test, which is also using this check, because for zarr they are data vars. Does it hurt the netcdf test to remove the lat/lons here?

gmertes (commented):
Thanks for this! I'll defer to others with more knowledge of CF conventions to review that part. We may also want to present this at ATS just for visibility, there may be other users relying on the current behaviour.

src/anemoi/inference/testing/checks.py
mpvginde: I was testing without zarr installed. So didn't notice that it was breaking the zarr-out. 
I will revert the `checks.py` file.

Probably in the zarr output the lat/lons can also be flagged as coordinates. I will have a look at this, but for this PR I agree it is best to revert.

gmertes:
Note to maintainers: let's do a release before merging.

src/anemoi/inference/outputs/netcdf.py
```diff
@@ -122,37 +127,59 @@ def open(self, state: State) -> None:
...
         compression = {}  # dict(zlib=False, complevel=0)
 
-        values = len(state["latitudes"])
-        self.reference_date = state["date"]
+        n_values = len(state["latitudes"])
 
-        if reference_date := getattr(self.context, "reference_date", None):
-            self.reference_date = reference_date
+        # start date of the forecast
+        reference_date = getattr(self.context, "reference_date", None) or state["date"]
```
mpvginde, line 133: @gmertes , I guess here the wrong reference_date can be set if the context doesn't contain `reference_date` (when does this happen?) and `write_initial_state=False`. I guess in that case `state["date"]` contains the valid time of the first timestep?

gmertes: More tech debt 🥰

The getattr can be replaced with a simple `self.context.reference_date`: after the recent Runner refactor, the context (runner) will always have the reference_date attribute ([see here](https://github.com/ecmwf/anemoi-inference/blob/eaeb36fae30c754a03b8dc0442fc56373b6bedef/src/anemoi/inference/runner.py#L112-L113)), but it can be None. 

It will be None when no date is set in the config, and then we rely on the input to find the date in the input file. The input will attach the date it found to the state, and indeed that's where we then find it. 

The logic is messy and could be improved, but I don't think there is a scenario where the reference date can be wrong or diverge.

mpvginde: Ok, makes sense now. See some thoughts on improving here: #551

src/anemoi/inference/outputs/netcdf.py
```diff
@@ -122,37 +127,59 @@ def open(self, state: State) -> None:
...
-            self.time_var.long_name = "time"
-            self.time_var.calendar = "gregorian"
+        # start time of forecast
+        # no dimensions, scalar variable
+        with LOCK:
+            self.reference_date_var = self.ncfile.createVariable("forecast_reference_time", "i8")
+            self.reference_date_var.standard_name = "forecast_reference_time"
+            self.reference_date_var.long_name = "start time of forecast"
+            self.reference_date_var.units = f"seconds since {EPOCH}"
+            self.reference_date_var.calender = CALENDAR
```
aaron-hopkinson: Typo
```suggestion
            self.reference_date_var.calendar = CALENDAR
```

src/anemoi/inference/outputs/netcdf.py
```diff
@@ -122,37 +127,59 @@ def open(self, state: State) -> None:
...
+        with LOCK:
+            self.time_var = self.ncfile.createVariable("time", "i8", ("time",), **compression)
+            self.time_var.standard_name = "time"
+            self.time_var.long_name = "valid time"
+            self.time_var.units = f"seconds since {EPOCH}"
+            self.time_var.calendar = CALENDAR
+            self.time_var.axis = "T"
+
+        # forecast period / lead time
+        # time dimension auxilary coordinate
```
aaron-hopkinson:
```suggestion
        # time dimension auxiliary coordinate
```

src/anemoi/inference/outputs/netcdf.py
```diff
@@ -122,37 +127,59 @@ def open(self, state: State) -> None:
...
+        with LOCK:
+            self.period_var = self.ncfile.createVariable("forecast_period", "i8", ("time",), **compression)
+            self.period_var.standard_name = "forecast_period"
+            self.period_var.long_name = "lead time"
+            self.period_var.units = "seconds"
 
+        # latitude / longitude
+        # values dimension auxilary coordinates
```
aaron-hopkinson:
```suggestion
        # values dimension auxiliary coordinates
```

aaron-hopkinson (changes requested):
A few minor changes (typos).

I think we need to include a global attribute that states the version of the conventions that the data adheres to, see:
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#identification-of-conventions

aaron-hopkinson:
@mpvginde Just out of interest, have you tried passing data outputted from this change through https://github.com/cedadev/cf-checker? That would be a good way of checking for CF compliance.

HCookie (commented):
Thanks for looking into this, could you add a test assert for the cf compliance, potentially using that tool Aaron mentioned?

aaron-hopkinson (approved):

PR #550, opened 2026-07-14
fix(temporal downscaler): do not squeeze time dimension

### LINKED ISSUE #544: Incorrect output tensor squeeze in temporal downscaler
### Location

`anemoi/inference/runners/temporal_downscaler.py` (inside `forecast()`, in the loop over temporal downscaling steps).

### Root cause

The model output tensor has shape `(batch, time, ensemble, grid, vars)` — 5D. The temporal downscaler runner was squeezing this down to `(time, grid, vars)` using a naive `np.squeeze()`, which removes all size-1 dimensions. This caused two distinct failures depending on how the squeeze was applied:

**Bug 1 — `IndexError: tuple index out of range`**

With `input_timestep=12H`, `output_timestep=6H`, and both boundaries excluded (`output_left_boundary: false`, `output_right_boundary: false`), the model produces exactly 1 intermediate output step. This means the `time` dimension has size 1. `np.squeeze()` collapsed it along with `batch` and `ensemble`, leaving a 2D tensor `(grid, vars)`. The subsequent `output.shape[2]` then failed with an `IndexError` since the tensor only had 2 dimensions.

**Bug 2 — `KeyError: 88`**

After fixing Bug 1 by specifying `axis=0` in the squeeze (removing only the batch dimension), the ensemble dimension (also size 1) was left in place. This produced a 4D tensor `(time=1, ensemble=1, grid, vars)` instead of the expected 3D `(time, grid, vars)`. Consequently, `output.shape[2]` returned the grid size (~13,000+) rather than the number of variables (88). The inner loop then iterated over grid points as if they were variable indices, failing at index 88 — the first index not present in the `output_tensor_index_to_variable` mapping (which only has 88 entries, keys 0–87).

Both bugs stem from the same line: the temporal downscaler runner's own squeeze logic that does not account for the ensemble dimension, unlike the standard runner in `runner.py`.

**Scope:** These bugs only manifest when the model has exactly 1 intermediate output step (`time=1`). With 2 or more intermediate steps (`time≥2`), `np.squeeze()` does not collapse the time dimension, so the code happens to produce the correct shape. For example, a model trained with `input_timestep=18H`, `output_timestep=6H` (2 intermediate steps) is not affected by this bug.

### Standard runner (correct pattern)

The standard runner in `anemoi/inference/runner.py` handles this correctly:

```python
if ndim == 4:
    # pre-multistep models output (batch, ensemble, values, variables)
    # add a time dimension of 1 for backwards compatibility
    tensor = tensor.unsqueeze(1)

outputs[dataset] = torch.squeeze(tensor, dim=(0, 2))  # shape: (time, values, variables)
```

It explicitly squeezes only the batch dimension (axis 0) and the ensemble dimension (axis 2), regardless of their sizes, and handles the 4D legacy case by inserting a time dimension first.

### Suggested fix

Replace the temporal downscaler runner's squeeze to mirror the standard runner:

```python
# Before (original — collapses time dim when time=1):
output = np.squeeze(y_pred[dataset].cpu().numpy())  # shape: (values, variables)

# After (correct — squeezes only batch and ensemble):
tensor = y_pred[dataset]
if tensor.ndim == 4:
    tensor = tensor.unsqueeze(1)
output = torch.squeeze(tensor, dim=(0, 2)).cpu().numpy()  # shape: (steps, values, variables)
```

Run a temporal downscaler with a single intermediate output step. A model trained with 2 or more intermediate steps is not affected by this bug.

### Version

v0.11.1

Cray Shasta, SUSE Linux Enterprise Server 15 SP6, AArch64

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

The Alan Turing Institute

VeraChristina:
## Description
Fix the temporal downscaler runner to work with checkpoints for which the time dimension (`steps`) in the output is 1. I have confirmed it runs a dummy checkpoint created with:
```
  input_timestep: 6h
  output_timestep: 3h
  output_right_boundary: False
```

https://github.com/ecmwf/anemoi-inference/issues/544

### DIFF
--- a/src/anemoi/inference/runners/temporal_downscaler.py
+++ b/src/anemoi/inference/runners/temporal_downscaler.py
@@ -321,7 +321,10 @@ class TemporalDownscalerMultiOutRunner(Runner):
 
                     # Detach tensor and squeeze (should we detach here?)
                     with ProfilingLabel("Sending output to cpu", self.use_profiler):
-                        output = np.squeeze(y_pred[dataset].cpu().numpy())  # shape: (values, variables)
+                        # squeeze batch and ensemble dims of tensor with shape: (batch, steps, ensemble, values, variables)
+                        output = np.squeeze(
+                            y_pred[dataset].cpu().numpy(), axis=(0, 2)
+                        )  # shape: (steps, values, variables)
 
                     if handler.trace:
                         handler.trace.write_output_tensor(


gmertes (approved):
@mc4117  ?

src/anemoi/inference/runners/temporal_downscaler.py
```diff
@@ -321,7 +321,10 @@ def forecast(
 
                     # Detach tensor and squeeze (should we detach here?)
                     with ProfilingLabel("Sending output to cpu", self.use_profiler):
-                        output = np.squeeze(y_pred[dataset].cpu().numpy())  # shape: (values, variables)
+                        # squeeze batch and ensemble dims of tensor with shape: (batch, steps, ensemble, values, variables)
+                        output = np.squeeze(
+                            y_pred[dataset].cpu().numpy(), axis=(0, 2)
```
mc4117, line 326: what happens if you have multiple output steps? The shape of the output has now changed?

mc4117 (approved):
LGTM! Thanks for implementing

PR #554, opened 2026-07-28
feat(inputs): Add opendap input source

HCookie:
## Description
Add opendap input source using earthkit data source.

```yaml
input: 
	opendap: URL_GOES_HERE_{date}
```

Allows for integrations with opendap and thredds servers

----
📚 Documentation preview 📚: https://anemoi-inference--554.org.readthedocs.build/en/554/

### DIFF
--- a/src/anemoi/inference/inputs/opendap.py
+++ b/src/anemoi/inference/inputs/opendap.py
@@ -51,7 +51,11 @@ class OpenDAPInput(EkdInput):
         url: str | list[str]
             The URL or list of URL's of the OpenDAP file / server.
             Python format strings can be used to dynamically construct the URL, with the following variables available:
-            - {date} : The date for which the input is being created.
+            - {date:...} : The date for which the input is being created.
+
+            It is expected that a user formats any variable correctly for the OpenDAP server, e.g. using strftime format codes for dates.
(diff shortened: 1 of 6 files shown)


src/anemoi/inference/inputs/opendap.py
```diff
@@ -0,0 +1,135 @@
...
+        super().__init__(context, metadata, **kwargs)
+        self.url = url if isinstance(url, list) else [url]
+
+    def _resolve_url(self, date: Date | None) -> list[str]:
+        """Resolve the URL for the given date."""
+        if date is None:
+            if any("{date" in u for u in self.url):
+                raise ValueError("Date must be provided to resolve the URL.")
+            return self.url
+        return [u.format(date=date) for u in self.url]
```
gmertes, line 69: I guess here you're relying on the default string format of a datetime object? So that's assuming the server has files with that exact format. Should we explicitly strftime it and maybe provide the option to change the format string?

HCookie: Here is an example of a thredds I was targeting, given it's possible for different servers to be layed out very differently, I thought it best to leave it to the user to define, as they may need it in many places
https://thredds.met.no/thredds/catalog/meps25epsarchive/2026/06/30/catalog.html?dataset=meps25epsarchive/2026/06/30/meps_det_vc_2_5km_20260630T21Z.nc

gmertes (approved):

PR #555, opened 2026-07-29
feat(inputs): add `forcings_from_forecast` option

HCookie:
## Description
Allow for forecast states to be used as forcings, by retrieving on step and base date instead of at step 0 for all valid dates.
Useful for running inference in a semi coupled way

```yaml
input:
	mars:
 		forcings_from_forecast: true
```

## TODO

- [x] Consider if temporal downscaler selection needs to be reworked 
	- A different probkem
- [x] Identify if other inputs are to be updated
	- All seem good
- [x] Unify dataset kwarg?
	- Slightly different, as the ds is a different shape

### DIFF
--- a/src/anemoi/inference/inputs/mars.py
+++ b/src/anemoi/inference/inputs/mars.py
@@ -339,7 +339,7 @@ class MarsInput(GribInput):
             The loaded forcings state.
         """
         if self.forcings_from_forecast:
-            LOG.debug("MarsInput: Loading forcings from forecast for dates: %s", dates)
+            LOG.debug("%s: Loading forcings from forecast for dates: %s", self.__class__.__name__, dates)
             date, steps = convert_dates_to_base_and_step(dates)
             retrieved_state = self.retrieve(self.variables, [date], step=steps)
         else:
--- a/src/anemoi/inference/inputs/mars.py
+++ b/src/anemoi/inference/inputs/mars.py
@@ -22,6 +22,7 @@ from anemoi.inference.types import State
 
 from . import input_registry
 from .grib import GribInput
+from .utils import convert_dates_to_base_and_step
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


mc4117:
Wouldn't it make sense to unify it in the temporal downscaler at the same time?

src/anemoi/inference/inputs/utils.py
```diff
@@ -0,0 +1,35 @@
...
+    -------
+    tuple[Date, list[int]]
+        The base date and the list of steps in hours.
+    """
+    if not dates:
+        raise ValueError("The list of dates is empty.")
+    datetimes = [to_datetime(date) for date in dates]
+
+    base_date = min(datetimes)
+    steps = [(dt - to_datetime(base_date)).total_seconds() // 3600 for dt in datetimes]
```
mc4117, line 33: this returns a float not an int right? I guess this could cause problems with mars requests

HCookie:
@mc4117 Could unify it such that this https://github.com/ecmwf/anemoi-inference/blob/9ce3fae7d5fcda0e633af3a1ad84d90542fb863a/src/anemoi/inference/runners/temporal_downscaler.py#L83-L97 uses the util, but I felt the rest was too different.

src/anemoi/inference/inputs/mars.py
```diff
@@ -22,6 +22,7 @@
 
 from . import input_registry
 from .grib import GribInput
+from .utils import convert_dates_to_base_and_step
```
mc4117, line 26: forcings_from_forecast does not seem to be added to the MarsInput.__init__ signature unlike for FDB and CDS. I guess this is passed through kwargs but there is a risk this could silently fail

mc4117:
Could be useful to add some unit tests for these features?

src/anemoi/inference/inputs/utils.py
HCookie:
```suggestion
    steps = [int((dt - to_datetime(base_date)).total_seconds() // 3600) for dt in datetimes]
```

src/anemoi/inference/inputs/mars.py
HCookie: Good catch

PR #558, opened 2026-07-31
test: Run pytest in parallel

gmertes:
## Description
The tests are starting to become a bit slow, we can run them in parallel to get a free speedup.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,7 @@ optional-dependencies.tests = [
   "hypothesis",
   "pytest",
   "pytest-mock",
+  "pytest-xdist",
 ]
 optional-dependencies.zarr = [ "zarr" ]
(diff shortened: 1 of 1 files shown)


gmertes:
This speeds things up locally and on Github runners, but seems to slow things down significantly on the self-hosted runners (downstream-ci). Need to investigate why. @mcocdawc any ideas?

HCookie (approved):
Looks good to me, so long as the slow down on the self hosted is resolvable on that side

mcocdawc:
I would not be surprised, if the slowdown on the selfhosted runners comes from overcrowding with processes.
I hope this week we will be already roll out the cubernetes cluster. This one will have differently sized runner groups, among them a `very-large` one. Could you wait until this is there and then run on this one?

mcocdawc:
Trying it in #581

PR #503, opened 2026-05-18
feat: flexible forecaster [experimental]

dietervdb-meteo:
## Description
This is a draft PR that adds to inference support for the flexible forecaster https://github.com/ecmwf/anemoi-core/pull/1070

This code serves two purposes: 1. be able to run experimental inference with a flexible forecaster. 2. highlight some code changes the would be required in inference if we generalize the forecaster.

Note that this is a draft PR. It might very well be the case that this code will not [from this branch, or as is] in the end be merged to main. Still I set up this PR to provide some visibility and to allow for easy comparison the changes introduced wrt main.

Example config to use the new runner:
```yaml
runner:
  flexible:
    input_offsets: ["-6H", "0H"]
    output_offsets: ["1H", "2H", "3H", "4H", "5H", "6H"]
    step_shift: "6H"
```

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

PR #560, opened 2026-08-04
feat: support flexible timestepping

dietervdb-meteo:
## Description
Introduces support for the more general Forecaster with flexible time steps. https://github.com/ecmwf/anemoi-core/pull/1070

The key change is that combining input and output into the next step's input is now handled using an index mapping stored in `advance_map`. For checkpoints post #1070 this is read directly from the inference metadata, for older checkpoints this is constructed out of other metadata pieces.

##  Additional notes ##
This PR is an alternative to https://github.com/ecmwf/anemoi-inference/pull/503, without a separare runner.


HCookie:
As github.com/ecmwf/anemoi-core/pull/1070 is still not merged, I'm going to move this PR into draft, but thanks for opening it, so we can discuss it.

src/anemoi/inference/tensors.py
```diff
@@ -498,10 +499,10 @@ def add_boundary_forcings_to_input_tensor(
             )  # shape: (1, dates, 1, values, variables)
             forcings = torch.from_numpy(forcings).to(self.context.device)  # Copy to device
 
-            for i in range(min(self.metadata.multi_step_output, self.metadata.multi_step_input)):
-                total_mask = np.ix_([0], [-(i + 1)], source.spatial_mask, source.variables_mask)
+            for out_idx, new_idx in self.context.checkpoint.advance_map["outin"]:
+                total_mask = np.ix_([0], [new_idx], source.spatial_mask, source.variables_mask)
                 input_tensor_torch[total_mask] = forcings[
-                    :, -(i + 1), ...
+                    :, out_idx, ...
```
gmertes: Here I would not get the advanced map from `self.context.checkpoint`, but instead you can get it via the `self.metadata`.

And then I think the `advanced_map` attribute under checkpoint becomes unused, and could be removed?

src/anemoi/inference/outputs/zarr.py
```diff
@@ -197,21 +198,25 @@ def open(self, state: State) -> None:
...
+                time += 1
         time += int(self.write_step_zero)
 
         if reference_date := getattr(self.context, "reference_date", None):
             self.reference_date = reference_date
 
-        if not self.write_step_zero and time_step is not None:
-            self.reference_date -= time_step
+        if not self.write_step_zero:
+            self.reference_date -= self.context.checkpoint.output_offsets[0]
```
gmertes: Same comment here as in the tensorhandler: you can go via `self.metadata`

src/anemoi/inference/metadata.py
```diff
@@ -231,6 +231,28 @@ def lagged(self) -> list[datetime.timedelta]:
...
+        result = [(s + 1) * self.timestep for s in result]
+        return sorted(result)
+
+    @cached_property
+    def rollout_shift(self) -> datetime.timedelta:
+        return self.output_offsets[-1]
+
+    @cached_property
+    def advance_map(self) -> Any:
+        """Index mapping to advance input and output into the next input"""
```
gmertes: Can you add a comment to say that this is used only by the rollout?

src/anemoi/inference/tensors.py
dietervdb-meteo: Ok, implemented. Now `advance_map` in the checkpoint class is nowhere called. Shall we remove the attribute from the code, or keep it for possible future use?

src/anemoi/inference/outputs/zarr.py
dietervdb-meteo: done

src/anemoi/inference/metadata.py
dietervdb-meteo: Updated the docstring, ok?

src/anemoi/inference/tensors.py
gmertes: I think we can just remove it, can always add it back if we ever need it.

src/anemoi/inference/inputs/dataset.py
```diff
@@ -292,8 +292,7 @@ def _load_basedates(self, basedates: list[Date]) -> Any:
         else:
             diff = idx[1] - idx[0]
             if not all(i == diff for i in np.diff(idx)):
-                # TODO: remove that restriction
-                raise ValueError("Dates do not have the same frequency")
+                return np.stack([self.ds[i] for i in idx], axis=0)
             s = slice(idx[0], idx[-1] + 1, diff)
 
         return self.ds[s]
```
HCookie, line 298: Given that it is now possible for flexible times to be got, I don't think this needs to be as complex anymore.

src/anemoi/inference/checkpoint.py
```diff
@@ -241,6 +241,40 @@ def target_explicit_times(self) -> Any:
         """Get the target explicit times."""
         return self._metadata.target_explicit_times
 
+    @property
+    def advance_map(self) -> Any:
```
HCookie: The `Any` type here is a bit too broad, surely it can be constrained?

src/anemoi/inference/checkpoint.py
```diff
@@ -241,6 +241,40 @@ def target_explicit_times(self) -> Any:
...
+
+        inin_dst = [dst for _, dst in inin]
+        assert inin_dst == sorted(inin_dst), f"{msg} inin not ordered by destination: {inin_dst}"
+
+        outin_dst = [dst for _, dst in outin]
+        assert sorted(inin_dst + outin_dst) == list(
+            range(n)
+        ), f"{msg} Inconsistent destinations missing. {advance_map}."
+
+        return advance_map
```
HCookie: This seems to just check the construction from the `metadata.py` class without any other information, why is the validation here?

src/anemoi/inference/metadata.py
```diff
@@ -231,6 +231,28 @@ def lagged(self) -> list[datetime.timedelta]:
...
+
+        result = [(s + 1) * self.timestep for s in result]
+        return sorted(result)
+
+    @cached_property
+    def rollout_shift(self) -> datetime.timedelta:
+        return self.output_offsets[-1]
+
+    @cached_property
+    def advance_map(self) -> Any:
```
HCookie: Broad type here too

src/anemoi/inference/runner.py
```diff
@@ -393,23 +393,22 @@ def forecast_stepper(
         is_last_step : bool
             True if it's the last step of the forecast.
         """
-        output_horizon = self.checkpoint.timestep * self.checkpoint.multi_step_output
-        steps = math.ceil(lead_time / output_horizon)
+        steps = math.ceil(lead_time / self.checkpoint.rollout_shift)
 
         LOG.info(
-            "Lead time: %s, time stepping: %s, Forecasting %s steps through %s autoregressive steps of %s prediction(s) each.",
+            "Lead_time=%s, step_shift=%s, output_offsets=%s, Forecasting %s steps through %s autoregressive steps of %s prediction(s) each.",
```
HCookie: I'm getting worried about the complexity of easily passing this log message. Could we have some variants based on verbosity?

src/anemoi/inference/runner.py
```diff
@@ -475,7 +474,7 @@ def forecast(
                 for d in dates:
                     dates_str += f"{d}, "
                 dates_str = f"{dates_str[:-2]})"
-                title = f"Forecasting, model call {s+1}: horizon {step}, freq. {self.checkpoint.timestep} {dates_str}"
+                title = f"Forecasting, model call {s+1}: horizon {step}, {dates_str}"
```
HCookie: Same concern with the complexity of this one, ultimately, a user just wants to know what time is being forecast,

src/anemoi/inference/tensors.py
dietervdb-meteo: Done.

src/anemoi/inference/checkpoint.py
dietervdb-meteo: Resolved, since modifications to `checkpoint.py` were reverted.

src/anemoi/inference/checkpoint.py
dietervdb-meteo: Yeah, was a bit of overkill. In any case, `advance_map` has now been removed from `checkpoint.py` since it is currently called only through metadata.

src/anemoi/inference/metadata.py
dietervdb-meteo: Fixed: [ba0a7d3](https://github.com/ecmwf/anemoi-inference/pull/560/commits/ba0a7d3d1d77a955d2435176b4f06f3801e66732)

src/anemoi/inference/runner.py
dietervdb-meteo: Moved this to `verbosity>0` since it is just a confirmation of how the stepping was set up (which is actually a property of the checkpoint).

src/anemoi/inference/runner.py
dietervdb-meteo: Changed this to short message on default and longer message upon `verbosity > 0`

src/anemoi/inference/inputs/dataset.py
dietervdb-meteo: We could indeed simply do
```python 
def _load_basedates(self, basedates: list[Date]) -> Any:
    idx = [self._find_index_for_date(d) for d in basedates]
    return np.stack([self.ds[i] for i in idx], axis=0)
```
But this is a bit less performant for cases where one does have contiguous dates, since the slice path is a single request instead of N. For a zarr-backed dataset that could have a latency impact? Of course we are not really envisioning models with 10k inputs, so maybe this is irrelevant and slightly simpler code is to be preferred?

dietervdb-meteo:
@japols , the new rollout code of this PR overwrites your changes of #566 . As far as I'm aware it should be as performant as what you had, but maybe this should be tested? In particular for the OOM issue you refer to in #566 ?

src/anemoi/inference/tensors.py
```diff
@@ -409,13 +409,11 @@ def copy_prognostic_fields_to_input_tensor(
 
         prognostic_fields = torch.index_select(y_pred, dim=-1, index=pmask_out)
 
-        # roll the input tensor in place to make room for the new prognostic fields
-        keep_steps = min(self.metadata.multi_step_output, self.metadata.multi_step_input)
-        for i in range(input_tensor_torch.shape[1] - keep_steps):
-            input_tensor_torch[:, i].copy_(input_tensor_torch[:, i + keep_steps])
+        for old_idx, new_idx in self.metadata.advance_map["inin"]:
+            input_tensor_torch[:, new_idx, :, :] = input_tensor_torch[:, old_idx, :, :]
```
dietervdb-meteo, line 413: HI @japols , here abouts are the key changes.

japols: thanks! assuming new_idx and old_idx are disjoint this should replace the old logic with an in-place overwrite, perfect :)

dietervdb-meteo: yes, that they are disjoint (and well ordered) is validated while training a new checkpoint or upon construnction for old checkpoints.

japols:
> @japols , the new rollout code of this PR overwrites your changes of #566 . As far as I'm aware it should be as performant as what you had, but maybe this should be tested? In particular for the OOM issue you refer to in #566 ?

I did check, this is indeed the case: same peak memory compared to main and no additional copies of the input tensor, nice work:
Screenshot 2026-08-31 at 14 02 18

HCookie (approved):
Thanks for your work on this, I will look at porting some of this to the metadata package in due time

PR #565, opened 2026-08-17
test(integration/meteoswiss): add varda-forecaster-det-sgm

### LINKED ISSUE #508: Fix and update MeteoSwiss integration test
eccodes local definition update borked the integration test, and the test is also no longer representative of the practical setup. We temporarily disabled it, it needs to be fixed and updated.

frazane:
Adds an integration test for the MeteoSwiss Varda stretched-grid deterministic forecaster. Closes #508.


HCookie (approved):
Thanks for fixing this.

PR #569, opened 2026-08-26
perf(inputs): keep inputs grid-sharded

japols:
## Description
Keep input and autoregressive rollout tensors sharded across model ranks, gathering forecast fields only to rank 0 when producing output.

Companion anemoi-models change: https://github.com/ecmwf/anemoi-core/pull/1345

For high-resolution tasks like 4km downscaling, storing full non-grid-sharded inputs can take up a majority of GPU memory. This PR makes the memory usage for parallel inference scalable i.e. n gpus/model -> 1/n x memory requirements for storing input tensors.

> [!WARNING]
> Requires this anemoi-core change for compatibility: https://github.com/ecmwf/anemoi-core/pull/1345

### DIFF
--- a/src/anemoi/inference/runner.py
+++ b/src/anemoi/inference/runner.py
@@ -274,6 +274,20 @@ class Runner(Context):
                             state[name]["fields"][field_name] = field.cpu().numpy()
             yield state
 
+    def prepare_forecast_input_tensors(
+        self, input_tensors_torch: dict[str, "torch.Tensor"]
+    ) -> dict[str, "torch.Tensor"]:
+        """Prepare device tensors before entering the autoregressive loop."""
+        return input_tensors_torch
+
+    def prepare_forecast_output_state(self, state: dict[str, State]) -> dict[str, State]:
+        """Prepare a forecast state for yielding without changing rollout state."""
+        return state
+
+    def grid_shard_slice(self, dataset: str) -> slice:
+        """Return the local interval along the grid dimension."""
+        return slice(None)
+
     @cached_property
     def autocast(self) -> Union["torch.dtype", str]:
         """The autocast precision."""
@@ -443,6 +457,7 @@ class Runner(Context):
                 dataset: torch.from_numpy(np.swapaxes(input_tensor_numpy, -2, -1)[np.newaxis, ...]).to(self.device)
                 for dataset, input_tensor_numpy in input_tensors_numpy.items()
             }
+            input_tensors_torch = self.prepare_forecast_input_tensors(input_tensors_torch)
 
             lead_time = to_timedelta(lead_time)
 
@@ -547,7 +562,7 @@ class Runner(Context):
 
                     # we only need to check the first dataset's step as they should all be the same
                     if next(iter(new_states.values()))["step"] <= lead_time:
-                        yield new_states
+                        yield self.prepare_forecast_output_state(new_states)
 
                 # No need to prepare next input tensor if we are at the last autoregressive step
                 if is_last_step:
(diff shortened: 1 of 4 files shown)


src/anemoi/inference/runner.py
```diff
@@ -696,14 +711,17 @@ def execute(self) -> None:
...
 
         for output in self.outputs.values():
             output.close()
 
+    def write_output_state(self, dataset: str, state: State) -> None:
+        """Apply top-level post-processors and write one forecast state."""
+        for processor in self.post_processors[dataset]:
+            state = processor.process(state)
+        self.outputs[dataset].write_state(state)
+
```
HCookie, line 726: This seems a change unrelated to this PR?

src/anemoi/inference/runners/parallel.py
```diff
@@ -30,6 +30,28 @@
...
+    padded = torch.empty(padded_shape, dtype=tensor.dtype, device=tensor.device)
+    padded[: tensor.shape[0]].copy_(tensor)
+
+    gathered = [torch.empty_like(padded) for _ in shard_sizes] if is_master else None
+    torch.distributed.gather(padded, gather_list=gathered, dst=0, group=process_group)
+
+    if not is_master:
+        return None
+
+    return torch.cat([shard[:size] for shard, size in zip(gathered, shard_sizes)], dim=0)
```
HCookie, line 52: This still materialises the full tensor on the master rank right? 
(just to check my understanding)

src/anemoi/inference/runner.py
```diff
@@ -443,6 +457,7 @@ def forecast(
                 dataset: torch.from_numpy(np.swapaxes(input_tensor_numpy, -2, -1)[np.newaxis, ...]).to(self.device)
                 for dataset, input_tensor_numpy in input_tensors_numpy.items()
             }
+            input_tensors_torch = self.prepare_forecast_input_tensors(input_tensors_torch)
```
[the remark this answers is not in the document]
japols: this is true but also a known limitation: afaik writing outputs in grib requires full fields (i.e. all gridpoints). a solution would be to either:
* chunk across the field/variable dimension (gather + write output in smaller chunks of fields)
* extend parallel output writing to all GPUs (instead of only spawning cpu writer tasks from rank 0) and perform an all-to-all operation to have full grids and sharded fields across GPUs that can then be written in parallel

thoughts @cathalobrien ? maybe good for a follow-up PR whenever needed

src/anemoi/inference/runners/parallel.py
```diff
@@ -179,6 +202,103 @@ def seed(self, comm_group: "torch.distributed.ProcessGroup | None") -> None:
...
+        assert process_group is not None
+        rank = process_group.rank()
+        start = sum(shard_sizes[:rank])
+        return slice(start, start + shard_sizes[rank])
+
+    def prepare_forecast_output_state(self, state: dict[str, Any]) -> dict[str, Any]:
+        """Gather all forecast fields together on rank 0 while keeping rollout state sharded."""
+        process_group = self.compute_client.process_group
+        if process_group is None:
+            return state
```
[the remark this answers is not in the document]
japols: follow-up PR potentially for @joherics to extend sharded tests to inference

src/anemoi/inference/runner.py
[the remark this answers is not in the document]
HCookie: It is possible an integration with `multio` may assist with the sharded writing, but for now, yes writing outputs in grib requires full fields.
Although this comment seems related to the input not the output tensor, which also materialises the full grid.

src/anemoi/inference/runners/parallel.py
japols: yeah, but only after the model has been applied so memory pressure shouldn't be too high. also see my comment above on potential improvements

src/anemoi/inference/runner.py
japols: I wanted to unify the Runner API to overwrite write_output_state for the ParallelRunner (which calls super().write_output_state only on the master rank)

src/anemoi/inference/tensors.py
```diff
@@ -499,10 +500,18 @@ def add_boundary_forcings_to_input_tensor(
             forcings = np.swapaxes(
                 forcings[np.newaxis, :, np.newaxis, ...], -2, -1
             )  # shape: (1, dates, 1, values, variables)
+
+            grid_slice = self.context.grid_shard_slice(self.dataset_name)
```
japols, line 502: @dietervdb-meteo could you review the sharding logic for the local spatial mask please?

PR #572, opened 2026-09-03
refactor(inputs/cutout): fix input state creation by removing coordinate loading

HCookie:
## Description
Removed fallback mechanism for loading coordinates from sources and adjusted input state creation.


HCookie:
So.....
The integration tests didn't run, and I missed that this broke the cosmo integration tests. 
`ekd` cannot get the lats/lons from the field, so it falls over to the checkpoint, which fails. ekd>=1.0 should fix it (tm)

PR #573, opened 2026-09-04
fix(metadata): correctly resolve checkpoint variable names in default_namer

### LINKED ISSUE #571: `default_namer` doesn't translate raw GRIB param/level back into the checkpoint's own variable name
### What happened?

Running inference crashes while writing the initial GRIB output state:

```python
Traceback (most recent call last):
File "/user-environment/bin/anemoi-inference", line 12, in <module>
sys.exit(main())
File ".../anemoi/inference/main.py", line 38, in main
cli_main(version, doc, COMMANDS)
File ".../anemoi/utils/cli.py", line 272, in cli_main
cmd.run(args)
File ".../anemoi/inference/commands/run.py", line 60, in run
runner.execute()
File ".../anemoi/inference/runner.py", line 695, in execute
self.outputs[dataset].write_initial_state(state)
File ".../anemoi/inference/outputs/tee.py", line 79, in write_initial_state
output.write_initial_state(state)
File ".../anemoi/inference/outputs/grib.py", line 269, in write_initial_state
variable = self.typed_variables[name]
~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'FI_100'
```

### Root cause

We use two separate naming conventions:

1. **The checkpoint's own internal variable names** — e.g. `z_100`, `z`.
2. **The raw GRIB data's own parameter names** — e.g. `FI`, `FIS`.

The checkpoint stores the mapping between the two itself. Each `variables_metadata` entry carries, alongside its own name (the dict key), a nested `mars: {param, levelist}` block recording that variable's raw-data identity:

```python
'z_100': {'mars': {'param': 'FI', 'levelist': 100, 'levtype': 'pl', ...}}
'z':     {'mars': {'param': 'FIS', 'levtype': 'sfc', ...}}
```

`Metadata.default_namer()` (src/anemoi/inference/metadata.py) is supposed to use that stored mapping to translate an incoming field raw param/levelist back into the checkpoint own variable name. Instead, it just concatenates the raw values directly: `f"{param}_{levelist}"`. For `param="FI"`, `levelist=100`, that produces `"FI_100"` and the subsequent lookup in `Metadata.typed_variables` (keyed by the checkpoint's own names) raises `KeyError`.

According to Claude, the bug in `default_namer()` has existed since the function was written, but stayed harmless because `wrap_state()` (post_processors/earthkit_state.py) used to hard-alias each field param metadata to the checkpoint's own name:

```python
# 0.11.1
super().__init__(name=field.name, param=field.name, ...)
```

so `default_namer` received `param="z_100"` directly, and its naive construction happened to return the correct answer. However, #539 (merged between 0.11.1 and 0.11.2) changed `wrap_state()` to expose the field's real mars metadata instead:

```python
# 0.11.2
metadata = dict(field._raw_metadata.get("mars", field._raw_metadata))
metadata.update(name=f, ...)
```

exposing `default_namer`'s bug, i.e. `param` is now `"FI"`, not a stand-in for `"z_100"`.

### Proposed fix

```python
def default_namer(self, *args: Any, **kwargs: Any) -> Callable[..., str]:
    assert len(args) == 0, args
    assert len(kwargs) == 0, kwargs

    # Reverse lookup from the checkpoint's own (param, levelist) to variable
    # name, using each variable's mars metadata. This lets us recover names
    # like `z_100` even when the input's GRIB `param` doesn't match the
    # checkpoint's variable name (e.g. `FI` vs `z`).
    param_levelist_to_name = {
        (variable.param, variable.level): name for name, variable in self.typed_variables.items()
    }

    def namer(field: ekd.Field, metadata: dict[str, Any]) -> str:
        param, levelist, levtype = (
            metadata.get("param"),
            metadata.get("levelist"),
            metadata.get("levtype"),
        )

        # Bug in eccodes that returns levelist for single level fields in GRIB2
        if levtype in ("sfc", "o2d"):
            levelist = None

        name = param_levelist_to_name.get((param, levelist))
        if name is not None:
            return name

        if levelist is None:
            return param

        return f"{param}_{levelist}"

    return namer
```

Take a checkpoint trained on data whose raw GRIB parameter names differ from the checkpoint's own internal variable names for at least one variable.

### Version

v0.11.2

Linux, x86_64 — SUSE Linux Enterprise Server 15 SP5 (kernel 5.14.21-150500.55.97_13.0.78-cray_shasta_c, HPE Cray/Shasta), CSCS Balfrin HPC cluster

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

MeteoSwiss

lclanzi:
## Description
Fixes GRIB field naming so it correctly maps back to the checkpoint's own variable names when the raw GRIB `param` doesn't match the checkpoint's internal naming (e.g. `FI` vs `z`).

Bugfix. `Metadata.default_namer()` built field names by concatenating the raw GRIB `param`/`levelist` directly (e.g. `FI_100`), instead of resolving them against the checkpoint's own variable metadata. This caused a `KeyError: 'FI_100'` crash whenever a checkpoint's variable name (e.g. `z_100`) didn't match the raw param it was retrieved under (e.g. `FI`).

This adds a `(param, levelist) -> name` lookup built from the checkpoint's own `typed_variables` mars metadata, used before falling back to the old naming behavior.

A unit test (`test_default_namer` in `tests/unit/test_metadata.py`) was added, covering a regular and a renamed case for both a surface and a pressure-level field. The renamed example variables were added to `testing/variables.py`.

Fixes #571

## Additional notes
No changes to the fallback behavior for fields with no matching checkpoint variable. The original `param`/`levelist` concatenation is still used in that case.

### DIFF
--- a/src/anemoi/inference/metadata.py
+++ b/src/anemoi/inference/metadata.py
@@ -602,10 +602,6 @@ class Metadata(LegacyMixin):
         assert len(args) == 0, args
         assert len(kwargs) == 0, kwargs
 
-        # Reverse lookup from the checkpoint's own (param, levelist) to variable
-        # name, using each variable's mars metadata. This lets us recover names
-        # like `z_100` even when the input's GRIB `param` doesn't match the
-        # checkpoint's variable name (e.g. `FI` vs `z`).
         param_levelist_to_name = {
             (variable.param, variable.level): name for name, variable in self.typed_variables.items()
         }
--- a/src/anemoi/inference/metadata.py
+++ b/src/anemoi/inference/metadata.py
@@ -602,9 +602,15 @@ class Metadata(LegacyMixin):
         assert len(args) == 0, args
         assert len(kwargs) == 0, kwargs
 
+        # Reverse lookup from the checkpoint's own (param, levelist) to variable
+        # name, using each variable's mars metadata. This lets us recover names
+        # like `z_100` even when the input's GRIB `param` doesn't match the
+        # checkpoint's variable name (e.g. `FI` vs `z`).
+        param_levelist_to_name = {
+            (variable.param, variable.level): name for name, variable in self.typed_variables.items()
+        }
+
         def namer(field: ekd.Field, metadata: dict[str, Any]) -> str:
-            # TODO: Return the `namer` used when building the dataset
-            warnings.warn("🚧  TEMPORARY CODE 🚧: Use the remapping in the metadata")
             param, levelist, levtype = (
                 metadata.get("param"),
                 metadata.get("levelist"),
@@ -614,6 +620,10 @@ class Metadata(LegacyMixin):
             # Bug in eccodes that returns levelist for single level fields in GRIB2
             if levtype in ("sfc", "o2d"):
                 levelist = None
+
+            name = param_levelist_to_name.get((param, levelist))
+            if name is not None:
+                return name
 
             if levelist is None:
                 return param
--- a/src/anemoi/inference/metadata.py
+++ b/src/anemoi/inference/metadata.py
@@ -611,7 +611,7 @@ class Metadata(LegacyMixin):
             else:
                 param_levelist_to_name[key] = name
         for key in ambiguous_keys:
-            del param_levelist_to_name[key]
+            param_levelist_to_name.pop(key, None)
 
         def namer(field: ekd.Field, metadata: dict[str, Any]) -> str:
             param, levelist, levtype = (
--- a/src/anemoi/inference/metadata.py
+++ b/src/anemoi/inference/metadata.py
@@ -602,9 +602,16 @@ class Metadata(LegacyMixin):
         assert len(args) == 0, args
         assert len(kwargs) == 0, kwargs
 
-        param_levelist_to_name = {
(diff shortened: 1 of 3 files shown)


src/anemoi/inference/metadata.py
```diff
@@ -602,9 +602,18 @@ def default_namer(self, *args: Any, **kwargs: Any) -> Callable[..., str]:
...
+        param_levelist_to_name: dict[tuple[Any, Any], str] = {}
+        ambiguous_keys: set[tuple[Any, Any]] = set()
+        for name, variable in self.typed_variables.items():
+            key = (variable.param, variable.level)
+            if key in param_levelist_to_name and param_levelist_to_name[key] != name:
+                ambiguous_keys.add(key)
+            else:
+                param_levelist_to_name[key] = name
+        for key in ambiguous_keys:
+            del param_levelist_to_name[key]
```
HCookie: I'm not sure `del` is the play here, how about `.pop`?

HCookie (commented):
Just one comment, but otherwise looks good

HCookie (approved):

PR #580, opened 2026-09-17
feat(outputs): Add exclusion option for variables in output 

maxineofficial:
## Description

Adds a new set of options for the `variables` config under `output`. This allows supported outputs (currently, `grib`, `netCDF`, `plot`, `printer`, `raw`, and `zarr`) to configure variables to keep or remove. 

This adds another configuration layer under `variable` which allows the user to `select` or `drop`. Only one option is allowed. The default behavior, where a single string or list indicates variables to keep, is still supported. 

If a user wanted to include all but one variable, they would have to input a long list of variables to keep. Now they can simply indicate they want to drop one. 

##  Additional notes ##

- I used `select` and `drop` to match datasets, but I can change it to `include` and `exclude` if preferred.
- This also removes the variable re-assignment in a few of the outputs which would skip the validation in the ABC's `__init__`. 
- The type of self.variables is changed to be a dictionary after `__init__` - I think this is only used internally for select/drop variables functionality.
- This also adds a ValueError check if the config is invalid, is this the correct way to handle this situation?

----
📚 Documentation preview 📚: https://anemoi-inference--580.org.readthedocs.build/en/580/


src/anemoi/inference/output.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         context: "Context",
         metadata: "Metadata",
         *,
-        variables: list[str] | None = None,
+        variables: list[str] | str | dict[str, list] | None = None,
```
HCookie:
```suggestion
        variables: list[str] | str | dict[Literal["select", "drop"], list] | None = None,
```
Given this is a constrained configuration, I think limiting the tying here would be good. Probably set the dict type as a global variable.
Don't think we need to go so far with a dataclass or anything.

src/anemoi/inference/output.py
```diff
@@ -64,28 +69,54 @@ def __init__(
         self._write_step_zero = write_initial_state
         self._output_frequency = output_frequency
 
-        self.variables = variables
-        if self.variables is not None:
-            if not isinstance(self.variables, (list, tuple)):
-                self.variables = [self.variables]
-
+        self.variables = self._validate_variables(variables) if variables is not None else None
```
HCookie: To remove some line complexity here, perhaps let `_validate_variables` return `None if variables is None`

src/anemoi/inference/output.py
```diff
@@ -64,28 +69,54 @@ def __init__(
...
+        variables : list[str] | str | dict[str, list]
+            Input variables.
+
+        Returns
+        -------
+        dict
+            A dictionary with one key, 'select' or 'drop', pointing to a list of variables:
+            {'select': [var1, var2...]}
+        """
+        # preserve existing functionality by defaulting to keeping the variables
```
HCookie: I prefer not to write docs / comments about the previous code

src/anemoi/inference/output.py
```diff
@@ -64,28 +69,54 @@ def __init__(
...
+        output_dict = {"select": variables} if not isinstance(variables, dict) else variables
```
HCookie: This mutates the type oddly, it could be possible for the dict to become `dict[str, str]`, which then downstream you have to contend with

src/anemoi/inference/output.py
```diff
@@ -64,28 +69,54 @@ def __init__(
...
+
+        if len(output_dict) > 1 and list(output_dict.keys())[0] not in ("select", "drop"):
+            raise ValueError(
+                f"Variables cannot include values other than a list, single value, or a one of `select` or `drop`. Found {output_dict.keys()}"
+            )
```
HCookie: What if dictionary has 0 keys? Plus there is no check if it has one key, but not select or drop

src/anemoi/inference/outputs/printer.py
```diff
@@ -164,7 +167,11 @@ def write_step(self, state: State) -> None:
         self.print()
         if self.metadata.multi_dataset:
             self.print(f"[{self.dataset_name}]", end=" ")
-        print_state(state, print=self.print, variables=self.variables, max_lines=self.max_lines)
+        keep_variables = self.variables.get("select") if self.variables is not None else None
+        drop_variables = self.variables.get("drop") if self.variables is not None else None
+        print_state(
+            state, print=self.print, variables=keep_variables, drop_variables=drop_variables, max_lines=self.max_lines
+        )
```
HCookie: As printer was one of the first outputs, I think it could make sense to rework it to be able to use the `skip_variables`? Could you explore the complexity of that?

src/anemoi/inference/output.py
```diff
@@ -64,28 +69,54 @@ def __init__(
...
         -------
         bool
             True if the variable should be skipped, False otherwise.
         """
-        return self.variables is not None and variable not in self.variables
+
+        if self.variables is None:
+            return False
+
+        skip_variable_select = "select" in self.variables.keys() and variable not in self.variables["select"]
```
HCookie:
```suggestion
        skip_variable_select = "select" in self.variables and variable not in self.variables["select"]
```
Don't need the `.keys()` here

tests/unit/outputs/test_printer.py
```diff
@@ -0,0 +1,41 @@
...
+
+@pytest.mark.parametrize(
+    "variables, expected_in_output, not_expected_in_output",
+    [
+        (None, ["z_500", "cp", "tp"], []),
+        ({"select": "cp"}, ["cp"], ["z_500", "tp"]),
+        ({"select": ["z_500", "cp"]}, ["cp", "z_500"], ["tp"]),
+        ({"drop": ["z", "cp"]}, ["tp", "z_500"], ["cp"]),
+        ({"drop": "cp"}, ["tp", "z_500"], ["cp"]),
+        ("all", ["z_500", "cp", "tp"], []),
```
HCookie: Also, could you use `pytest.Param` and id the tests?

docs/inference/configs/outputs.rst
```diff
@@ -18,13 +18,20 @@ The ``variables`` parameter allows you to subset the output to only include
...
 output, reducing output file size and processing time.
 
+You can either provide a list or single variable to include, or specify that you want to `select` or `drop` variables.
+Only one of "select" or "drop" should be provided.
+
+For `select`, only the variables provided in the list will be included in the output. For `drop`, only the
+variables NOT in the list will be included in the output.
+
 .. literalinclude:: yaml/outputs_variables.yaml
    :language: yaml
```
HCookie, line 30: Can you also add a drop example here?

HCookie (changes requested):
A couple of comments, but looks mostly good

> I used select and drop to match datasets, but I can change it to include and exclude if preferred.

Matching datasets make sense, happy to keep it like this.

> This also adds a ValueError check if the config is invalid, is this the correct way to handle this situation?

Yep, raising an informative exception with a ValueError is good

src/anemoi/inference/output.py
maxineofficial: Right now that is handled in the return call of this validation function (just makes sure all the vals in the dict are lists.) I can move it back up here if you prefer. It is slightly awkward in either case, unfortunately.

src/anemoi/inference/output.py
maxineofficial: I'll add a check around zero keys (plus a test case.) This should check if it has one key that isn't select or drop: `list(output_dict.keys())[0] not in ("select", "drop")`

src/anemoi/inference/outputs/printer.py
maxineofficial: Sure, I was looking at that initially, but ultimately this `print_state()` function isn't inside the class so it would require moving it inside of `PrinterOutput` and I wasn't sure if it was used for anything (nothing inside the code, but it's possibly used downstream by someone). If it's fine to move it then that's a pretty easy change.

maxineofficial: Or I could leave the function outside and add some code here to open up the state and assemble the variables before the `print_state` call, which would be very easy but duplicates a little code.

HCookie: I'd say bring it inside the class, it's a small enough function anyways

src/anemoi/inference/output.py
HCookie: How I wish python was more strongly typed.....

maxineofficial: Same 😔

src/anemoi/inference/output.py
```diff
@@ -26,6 +32,54 @@
 LOG = logging.getLogger(__name__)
 
 
+class OutputVariableConfig(BaseModel):
```
maxineofficial, line 35: Don't love this here, but couldn't find a better spot - suggestions welcome!

src/anemoi/inference/outputs/printer.py
```diff
@@ -45,7 +47,7 @@ def __init__(
         metadata: Metadata,
         *,
         path: Path | None = None,
-        variables: ListOrAll | None = None,
+        variables: ListOrAll | VARIABLE_CONFIG_TYPES = OutputVariableConfig(),
```
HCookie: Setting an instantiated object as the default, I think means it is shared across all instances of this class. Probably not a problem here, but not a standard good thing.

src/anemoi/inference/output.py
```diff
@@ -27,6 +32,54 @@
...
+    """Type for output variable configuration settings.
+
+    Only one of `select` or `drop` have values, with the other being []. `select` indicates that only the variables provided
```
HCookie: Do you check that only one is set?

src/anemoi/inference/output.py
```diff
@@ -27,6 +32,54 @@
 LOG = logging.getLogger(__name__)
 
 
+class OutputVariableConfig(BaseModel):
```
HCookie: Might be good to have a `.continue(var) -> bool` method on this object?

src/anemoi/inference/output.py
```diff
@@ -117,8 +143,8 @@ def skip_variable(self, variable: str) -> bool:
         if self.variables is None:
             return False
 
-        skip_variable_select = "select" in self.variables and variable not in self.variables["select"]
-        skip_variable_drop = "drop" in self.variables.keys() and variable in self.variables["drop"]
+        skip_variable_select = self.variables.select is not None and variable not in self.variables.select
+        skip_variable_drop = self.variables.drop is not None and variable in self.variables.drop
```
HCookie: If you add a method to the `OutputVariableConfig` this could become a simple forward method?

src/anemoi/inference/output.py
maxineofficial: Not when it's set in code rather than through the config - I added that.

src/anemoi/inference/outputs/printer.py
maxineofficial: You're right, fixed

src/anemoi/inference/output.py
maxineofficial: Not sure what you mean here - is this covered by the `skip(var) -> bool` method I just added?

src/anemoi/inference/outputs/printer.py
```diff
@@ -33,78 +33,6 @@
 ListOrAll = Union[list[str], Literal["all"]]
 
 
-def print_state(
```
gmertes: This is a bit hidden, but this function is used in our HF notebooks so I'd keep it public:
https://huggingface.co/ecmwf/aifs-single-2.0/blob/main/run_AIFS_v2.0.ipynb

src/anemoi/inference/types.py
```diff
@@ -40,3 +45,65 @@
...
+
+    def skip(self, variable: str) -> bool:
+        """Return True if the provided variable should be skipped, False otherwise."""
+        skip_variable_select = self.select is not None and variable not in self.select
+        skip_variable_drop = self.drop is not None and variable in self.drop
+
+        return skip_variable_select or skip_variable_drop
+
+
+OutputVariableConfigUnion = Union[OutputVariableConfig, str, list[str], dict[Literal["select", "drop"], str], None]
```
gmertes: Having the type here definitely makes sense, but not sure if having the schema here also makes sense. But I'm happy to think about that in a follow up.

gmertes (commented):
Very nice 👌

src/anemoi/inference/outputs/printer.py
HCookie: Given we make those, and it's easy enough to rework, I think it's fine to move it.

maxineofficial: Harrison and I did discussed this in person last week - happy to update HF if that would be helpful

HCookie (dismissed):
Thanks for your work on this. This will be quite a useful feature. Pinging @gmertes for a final review

gmertes (approved):
🚀

PR #581, opened 2026-09-18
ci: Run pytest in parallel on arc-runner-very-large

mcocdawc:
## Description

Runs the test suite with `pytest-xdist` on the `arc-runner-very-large` ARC scale set instead of GitHub's `ubuntu-latest`.

Two commits:

- `test: Run pytest in parallel` — adds `pytest-xdist` to the `tests` extra and sets `addopts = "-n auto --maxprocesses=16"`.
- `ci: Run pytest on arc-runner-very-large` — moves the `checks` job onto the big runner.

`-n auto` only pays off if there are cores to spread across, and `ubuntu-latest` has 2. The reusable `ecmwf/reusable-workflows/.github/workflows/qa-pytest-pyproject.yml@v2` hardcodes `runs-on: ubuntu-latest` and exposes no runner input, so the `checks` job is inlined here to set:

```yaml
runs-on: arc-runner-very-large
container:
  image: eccr.ecmwf.int/public-ci-images/ubuntu24.04-base:latest
```

That runner/image pairing follows `ecmwf/ci-infrastructure`'s `smoke-test-runners.yml` and the manifests in `ecmwf/ecbuild` (which use `arc-runner-normal`). The 3.11/3.12/3.13 matrix is unchanged; `uv` provisions each version since the image only ships 3.12.

No issue — exploratory, to see what the new runners do for CI wall time.

## Additional notes

This is a **trial balloon and should not be merged on a green tick alone** — the point is the numbers in the log, not the checkmark.

- A `Report runner sizing` step prints `nproc`, the CPU affinity mask and memory. Worth reading before anything else: under Kubernetes, `-n auto` can see the *node's* CPU count rather than the pod's limit, which is what `--maxprocesses=16` caps. If affinity comes back well below `nproc`, the cap wants retuning to match.
- Inlining the job means this repo no longer picks up fixes to the shared reusable workflow. If the runner turns out to be worth keeping, the better end state is a runner input on `qa-pytest-pyproject.yml` upstream rather than this copy.
- Unverified until CI actually runs: that `arc-runner-very-large` is enabled for this repo, and that `eccr.ecmwf.int` pulls anonymously from the cluster. Both are used without a registry login in the repos above, so I expect them to work, but neither is confirmed here.
- No timing baseline is captured in this PR. To get a real before/after, the `ubuntu-latest` job could be kept alongside for one run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mcocdawc:
## Timings

Three runs, same 298 tests. **Baseline** = last `main` run on `ubuntu-latest` without xdist ([34999873549](https://github.com/ecmwf/anemoi-inference/actions/runs/34999873549)); **run 1** = ARC + xdist ([35363913654](https://github.com/ecmwf/anemoi-inference/actions/runs/35363913654)); **run 2** = + CPU torch and workspace-local uv cache ([35367054568](https://github.com/ecmwf/anemoi-inference/actions/runs/35367054568)).

| | 3.11 | 3.12 | 3.13 |
|---|---|---|---|
| **whole job** — baseline | 250s | 255s | 245s |
| run 1 | 263s | 275s | 315s |
| run 2 | **122s** | **116s** | **108s** |
| **`pytest` step** — baseline | 208s | 226s | 198s |
| run 1 | 64s | 67s | 62s |
| run 2 | 51s | 61s | 54s |
| **`Install dependencies`** — baseline | 32s | 18s | 36s |
| run 1 | 163s | 176s | 225s |
| run 2 | 42s | 28s | 21s |

Slowest job: **255s → 116s, a 2.2x speedup** over the `ubuntu-latest` baseline.

Run 1 was *slower* than the baseline despite a 3.3x faster test step, because installation regressed 5-9x. uv's phase timings split that into two independent causes, both now fixed:

- **`Prepared` 2m13s-3m08s → 12.8-37.0s.** `torch` pulls the full CUDA stack by default — cudnn 595MB, torch 555MB, cublas 553MB, and ~17 more `nvidia-*`/`cuda-*` wheels, roughly 2.9GB that cannot execute on a CPU runner. It was invisible on hosted runners because the Azure link hid it; the cluster's egress to PyPI does not. `UV_TORCH_BACKEND=cpu` drops the resolution from 246 to 227 packages.
- **`Installed` 24.6-31.0s → 0.4-1.0s.** uv was warning `Failed to hardlink files; falling back to full copy` — ARC overlays the workspace per job, so the default cache sat on a different filesystem than `.venv` and all 227 packages were byte-copied. Pointing `UV_CACHE_DIR` into the workspace restores hardlinking; the warning is gone.

Outcomes are identical across all three runs and all three Python versions — `290 passed, 2 skipped, 6 xfailed` — so CPU-only torch changes nothing about what executes. Runner reports `nproc: 32`, affinity 32, 62GiB, so `--maxprocesses=16` is the binding cap rather than the pod.

### Caveats

- **Single sample per configuration.** The 3.11/3.12/3.13 spread within run 2 (108-122s) is comparable to some of the differences being read, and `Prepared` still varies 12.8-37.0s across legs, which is network jitter rather than anything structural.
- Inlining the job means this repo no longer tracks fixes to the shared `qa-pytest-pyproject.yml`. If this lands, the better end state is a runner input upstream instead of this copy.
- `UV_TORCH_BACKEND=cpu` is correct for CI only. It must not be copied anywhere that needs GPU torch.

